PR #5, opened 2018-04-10
Ch migrations scripting 156461261

JoyLubega:
#### What does this PR do?
This PR creates a migration script and cammands to be used in Migration.
#### Description
After creating or making changes in the  models, run the following commands;
`python manage.py db init`
`python manage.py db migrate`
`python manage.py db upgrade`
#### screenshots
screen shot 2018-04-10 at 11 42 49
screen shot 2018-04-10 at 11 42 06
screen shot 2018-04-10 at 11 42 32

#### Pivotal tracker Story
#156461261


config.py
```diff
@@ -12,18 +12,15 @@ def init_app(app):
 
 class DevelopmentConfig(Config):
     DEBUG = True
-    SQLALCHEMY_DATABASE_URI = os.environ.get('DEV_DATABASE_URL') or \
-    'sqlite:///' + os.path.join(basedir, 'dev-db.sqlite')
+    SQLALCHEMY_DATABASE_URI = 'postgresql://localhost/mrm_db'
```
PatrickCmd: I think we should not hard code DATABASE_URI variables. It's better to  use environment variables such as SQLALCHEMY_DATABASE_URI = os.getenv(DEV_DATABASE_URL).

JoyLubega: okay, Let me fix that

app.py
```diff
@@ -1,8 +1,11 @@
 from flask import Flask
+from flask_sqlalchemy import SQLAlchemy
```
shemogumbe: I would not encourage the use of flask-sqlalchemy but sqlalchemy itself as it has better compatibility with graphene-sqlalchemy for object type creation.

app.py
```diff
@@ -1,8 +1,11 @@
...
 
+#local imports
 from config import config
 
-
+#initializing a db instance
+db = SQLAlchemy()
```
shemogumbe: Let us go the route of using sqlalchemy and not flask-sqlalchemy

manage.py
```diff
@@ -1,12 +1,16 @@
 import os
 
 from flask_script import Manager, Shell
+from flask_migrate import Migrate, MigrateCommand
```
shemogumbe: I would prefer that we use alembic for migrations

alembic/env.py
```diff
@@ -0,0 +1,73 @@
...
+
+# Interpret the config file for Python logging.
+# This line sets up loggers basically.
+fileConfig(config.config_file_name)
+
+# add your model's MetaData object here
+# for 'autogenerate' support
+# from myapp import mymodel
+# target_metadata = mymodel.Base.metadata
+import os, sys
```
PatrickCmd: Accord to the pep8 code style guide, I think we should avoid **import os**, sys to
**import os**
**import sys**

alembic.ini
```diff
@@ -0,0 +1,76 @@
...
+# to alembic/versions.  When using multiple version
+# directories, initial revisions must be specified with --version-path
+# version_locations = %(here)s/bar %(here)s/bat alembic/versions
+
+# the output encoding used when revision files
+# are written from script.py.mako
+# output_encoding = utf-8
+
+#sqlalchemy.url = driver://user:pass@localhost/dbname
+sqlalchemy.url = postgresql://joylubega@localhost/mrm_db
```
PatrickCmd: I think it's better to use environment variables than hard coding variables.
So it would be better to have **sqlalchemy.url = os.getenv('some environment variable')**

shemogumbe: Unless we user environment variables configures at the env file, this will not be scalable as people trying to run migrations will need to only do so in your computer

JoyLubega: okay.. Thanks @shemogumbe . I have fixed it

JoyLubega:
How to run migrations
`Alembic upgrade head`
`Alembic revision --autogenerate`

PR #7, opened 2018-04-13
#[156466913] Automate deployment 

dkam26:
#### What does this PR do?
Automate creating db, running migrations and starting the app locally.
Clone ```https://github.com/andela/mrm_api.git``` to your local machine
Create a virtual environment and RUN ```pip install -r requirements.txt```
RUN  ```fab prepare_to_host:user=$username``` with $username equal to the system user


.gitingore
```diff
@@ -0,0 +1,4 @@
+.vscode/
+env/
+__pycache__/
+*.pyc
```
proxiex: Adding a new line will be great!

PR #11, opened 2018-04-23
#156614968  Update-room-info 

owenbob:
#### What does this PR do?
- This PR creates a mutation that allows a user to select a specific room by name and adjust it's details appropriately.
- Run the server 
- Test the mutation  at   **localhost:5000/mrm**
- #156614968 
 The screenshots take you through the key functionality for the PR. Here goes:

- Let's see what we have in our database.
image

- Now let us update all fields for a room with room_id `5`.
image

- Now let us try updating only a field we want. I this case lets update `room_type`
image

- If we query rooms, we see that the update has been made accordingly.
image


api/room/schema.py
```diff
@@ -32,20 +33,21 @@ def mutate(self, info, **kwargs):
...
         
 
     room = graphene.Field(Room)
     
-    def mutate(self, info,name, **kwargs):
+    def mutate(self, info, id, **kwargs):
+        validate_empty_fields(**kwargs)
         query_room = Room.get_query(info)
-        exact_room = query_room.filter(RoomModel.name == name).first()
+        exact_room = query_room.filter(RoomModel.id == id).first()
```
shemogumbe: LGTM

api/room/schema.py
```diff
@@ -32,20 +33,21 @@ def mutate(self, info, **kwargs):
...
 
-        if kwargs.get("new_name"):
-            exact_room.name = kwargs["new_name"]
+        if kwargs.get("name"):
```
shemogumbe: We could get cleaner code if we just spread the kwargs and supplied them as fields to the object, this can be achieved by supplying them to the save method in the format save(**kwargs), this can even be done in a helper class or we can override the save method of the object type.

fixtures/room/room_update_fixtures.py
```diff
@@ -0,0 +1,107 @@
+
+null = None
+
+query_update_all_fields = '''mutation{
+                updateRoom(name:"Entebbe",newName:"Jinja",capacity:8,roomType:"board room"){
```
shemogumbe: Avoid this name/new name logic, once id is supplied as a required positional argument, any other thing that is supplied is updated, any other field that is not supplied is left as was

fixtures/room/room_update_fixtures.py
```diff
@@ -0,0 +1,107 @@
...
+                    "room": {
+                        "name": "Jinja",
+                        "capacity": 8,
+                        "roomType": "board room"
+                    }
+                    }
+                }
+                }
+
+query_update_only_required_field = '''mutation{
```
shemogumbe: Avoid the name/new name logic

owenbob: Thanks. This has been resolved.

api/room/schema.py
```diff
@@ -29,12 +29,41 @@ def mutate(self, info, **kwargs):
...
+    class Arguments:
+        room_id = graphene.Int()
+        room_type = graphene.String()
+        capacity = graphene.Int()
+        name = graphene.String()
+        
+
+    room = graphene.Field(Room)
+    
+    def mutate(self, info, room_id, **kwargs):
```
owenbob: @shemogumbe. The changes requested for updating using the room_id have been implemented.

api/room/schema.py
owenbob: @shemogumbe, Thanks. I totally agree and moving forward code quality shall be upheld. I have refactored by introducing an `update_entity_fields` in the utilities. This method can be used for any other mutation for updating.

api/room/schema.py
```diff
@@ -29,12 +29,37 @@ def mutate(self, info, **kwargs):
...
+        validate_empty_fields(**kwargs)
+
+        query_room = Room.get_query(info)
+        exact_room = query_room.filter(RoomModel.id == room_id).first()
+        
+        update_entity_fields(exact_room,**kwargs)
```
shemogumbe: LGTM

fixtures/room/room_update_fixtures.py
shemogumbe: LGTM

PR #15, opened 2018-04-27
Ft get all rooms 156987495

owenbob:
#### What does this PR do?
- This PR ensures that a user is able to get all rooms.The queries a user can make are : 
```
        - Get all rooms
        - Get rooms in a block
        - Get rooms in a location
        - Get room by ID
        - GET all rooms following the room hierarchy in all locations
```
- Run the server
- #156987495
- Get all rooms
image

- Get rooms in a block
image

- Get rooms in a location
image

- Get room by ID
image

- GET all rooms following the room hierarchy in all locations
image


api/block/schema.py
```diff
@@ -0,0 +1,25 @@
...
+from graphene_sqlalchemy import (
+    SQLAlchemyObjectType, 
+    SQLAlchemyConnectionField
+)
+from api.block.models import Block as BlockModel
+
+class Block(SQLAlchemyObjectType):
+    class Meta:
+        model = BlockModel
+        interfaces = (relay.Node,)
```
shemogumbe: Avoid using relay Nodes as they add extra fields edges and Node to the query payload and response

api/floor/schema.py
```diff
@@ -0,0 +1,18 @@
...
+    SQLAlchemyConnectionField,
+    SQLAlchemyObjectType
+)
+
+from api.floor.models import Floor as FloorModel
+
+class Floor(SQLAlchemyObjectType):
+    class Meta:
+        model = FloorModel
+        interfaces = (relay.Node,)
```
shemogumbe: Same as comment above

tests/test_rooms/test_query_rooms.py
```diff
@@ -0,0 +1,16 @@
...
+from fixtures.room.room_fixtures import (
+    rooms_query,
+    query_rooms_response,
+    room_query_by_id,
+    room_query_by_id_response
+)
+class QueryRooms(BaseTestCase):
+    def test_query_rooms(self):
+        query_rooms = self.client.execute(rooms_query)
+        assert query_rooms == query_rooms_response
```
shemogumbe: I would prefer the use assertEquals instead of the equal sign, remember more words mean more clarity, less logic for the brain in order to make sense of the code

api/block/schema.py
owenbob: I totally agree. I have implemented the requested changes.

api/floor/schema.py
owenbob: Requested changes have been implemented

tests/test_rooms/test_query_rooms.py
owenbob: I have taken this into consideration and I agree. I have implemented the requested changes accordingly.

alembic/env.py
```diff
@@ -15,11 +15,17 @@
 # for 'autogenerate' support
 # from myapp import mymodel
 # target_metadata = mymodel.Base.metadata
+<<<<<<< HEAD
+import os
+import sys
+sys.path.append(os.getcwd())
+=======
```
PatrickCmd: I think you need to resolve these merge conflicts.

alembic/env.py
```diff
@@ -15,11 +15,17 @@
...
 # target_metadata = None
 import os
 import sys
 sys.path.append(os.getcwd())
 
+>>>>>>> 886bd3f4174947f2fb32aba10b8f635d69cbd92a
```
PatrickCmd: Conflict not taken care of.

owenbob: Thanks for noting this.

alembic/env.py
owenbob: Thanks, I will do that immediately.

api/location/schema.py
```diff
@@ -0,0 +1,39 @@
...
+from api.equipment.models import Equipment 
+
+class Location(SQLAlchemyObjectType):
+    class Meta:
+        model = LocationModel
+
+class Query(graphene.ObjectType):
+    node = relay.Node.Field()
+    all_locations = graphene.List(Location)
+    get_rooms_in_a_location =graphene.List(
```
PatrickCmd: I think this should be **get_location** not **get_rooms_in_a_location**

api/location/schema.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+    def resolve_all_locations(self,info):
+        query = Location.get_query(info)
+        return query.all()
+
+    def resolve_get_rooms_in_a_location(self,info,location_id):
+        query =Location.get_query(info)
+        result = query.filter(LocationModel.id == location_id)
+        return result
```
PatrickCmd: The resolver method does not return what the name suggests **get rooms in a location**, actually what it returns is location based on the **location_id**. So I thick it should be **resolve_get_location**

api/block/schema.py
```diff
@@ -0,0 +1,28 @@
...
+    )
+
+    def resolve_blocks(self,info):
+        query = Block.get_query(info)
+        return query.all()
+
+    def resolve_get_rooms_in_a_block(self,info,block_id):
+        query = Block.get_query(info)
+        result = query.filter(BlockModel.id == block_id)
+        return result
```
PatrickCmd: The resolver method does not return what the name suggests **get rooms in a block**, actually what it returns is location based on the **block_id**. So I thick it should be **resolve_get_block**

api/block/schema.py
```diff
@@ -0,0 +1,28 @@
...
+from api.block.models import Block as BlockModel
+
+class Block(SQLAlchemyObjectType):
+    class Meta:
+        model = BlockModel
+
+class Query(graphene.ObjectType):
+    node = relay.Node.Field()
+    floor = graphene.List(Block)
+    get_rooms_in_a_block = graphene.List(
```
PatrickCmd: I think this should be **get_block** not **get_rooms_in_a_block**

api/block/schema.py
owenbob: According to the flow on the front end, one needs to query rooms in a hierarchy. So the rooms_in_a_block will be from the block, the floor and finally rooms. Please refer to the screenshot.

api/location/schema.py
owenbob: According to the flow on the front end, one needs to query rooms in a hierarchy. So the rooms_in_a_location will be from the location, the block, the floor and finally rooms. Please refer to the screenshot.

owenbob: Thanks for that insight. However, according to the story requirements and the how the flow of the application goes on the front end, these room queries are meant to be hierarchical. That is why I went for this naming convention. Kindly take a look at the story on the PT board for more clarity.

api/block/schema.py
owenbob: Thanks. The flow is being followed as it necessitates by the front-end flow.

api/location/schema.py
owenbob: kindly see why this flow was followed  above.

PatrickCmd:
@owenbob I think your PR naming is following our git naming conventions and best practices as described here https://github.com/andela/bestpractices/wiki/Git-naming-conventions-and-best-practices

PR #27, opened 2018-05-21
 #156873054 add office

proxiex:
#### What does this PR do?
- create office model and schema
- add office model and schema
- write test cases for both schema and model
- modify location relationship, test files and fixtures
- modify block relationship, test files and fixtures

- Add an office to a location
- run app
- visit http://localhost:5000/mrm

n/a

[156873054](https://www.pivotaltracker.com/n/projects/2154921)

screen shot 2018-06-26 at 1 31 17 pm

#### Questions:
n/a


alembic/env.py
```diff
@@ -19,24 +19,18 @@
...
 
->>>>>>> 886bd3f4174947f2fb32aba10b8f635d69cbd92a
 from helpers.database import Base
 from api.location.models import Location
 from api.block.models import Block
 from api.floor.models import Floor
 from api.room.models import Room
 from api.room_resource.models import Resource
-
+from api.office.model import Office
```
PatrickCmd: I think office is the same as block, why introduce another model **office**

proxiex: I don't think they are the same thing. An office "EPIC Tower" for example could have block A, B, C etc.

PatrickCmd: Then the relationship should also change, location is related to office, **blocks = relationship('Block')** in **location model** changes to **offices = relationship('Office') ** and office is related to blocks, **location = relationship('Location')** changes to **blocks = relationship('Blocks')** in **office model**

proxiex: I guess it should. But will I be the one to do that?

PatrickCmd: Exactly yes. You are.

proxiex: 🤨

PR #37, opened 2018-06-08
#157912738  View all devices 

owenbob:
#### What does this PR do?
 - This PR  enables a user to create ,view and update devices.

- Run the server
- Run `createDevice` mutation
- Run `allDevices` query
- Run `updateDevice` mutation

- #157912738 
`createDevice` mutation

image

`allDevices` query

image

`updateDevice` mutation

image


api/location/schema.py
```diff
@@ -3,7 +3,7 @@
 
 from api.location.models import Location as LocationModel
 
-from api.room_resource.models import Resource as ResourceModel  # noqa: F401
+from api.devices.models import Devices as DevicesModel  # noqa: F401
```
PatrickCmd: I think devices model is not necessary in the locations schema.

api/devices/schema.py
```diff
@@ -0,0 +1,67 @@
...
+    device = graphene.Field(Devices)
+
+    def mutate(self, info, device_id, **kwargs):
+        validate_empty_fields(**kwargs)
+
+        query_device = Devices.get_query(info)
+        exact_device = query_device.filter(
+            DevicesModel.id == device_id
+            ).first()
+        exact_device.date_added = datetime.now()
```
shemogumbe: Before executing any logic with the device object, first check that it actually exists?

fixtures/devices/devices_fixtures.py
```diff
@@ -0,0 +1,88 @@
...
+                name
+                location
+                deviceType
+                }
+            }
+            }
+'''
+
+
+expected_update_device_response = {
```
shemogumbe: LGTM

api/location/schema.py
owenbob: It is as you are trying it is a declaration that you should nest till the `devices` model.

PR #39, opened 2018-06-11
[ft156874253]Add decode token to  return UserCredentials

JoyLubega:
#### What does this PR do?
Get the token from headers and Return User Credentials in the Payload.

#### How should this be manually tested?/
N/A
## #ft156874253

### DIFF
--- a/helpers/auth/token_authentication.py
+++ b/helpers/auth/token_authentication.py
@@ -1,0 +1,39 @@
+import jwt
+from flask import request, jsonify
+
+
+class Authentication:
+    """ Authenicate token and all role based authorization
+    :methods
+        verify
+        decode_token
+        user_roles
+        auth_required
+    """
+
+    def get_token(self):
+        token = request.headers.get('token')
+        return token
+
+    def decode_token(self):
+        """
+        Decodes the auth token
+        :param
+        :return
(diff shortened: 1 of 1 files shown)


helpers/auth/token_authentication.py
```diff
@@ -0,0 +1,39 @@
...
+        try:
+            auth_token = self.get_token()
+            payload = jwt.decode(auth_token, verify=False)
+            return payload['UserInfo']  # Return User Info
+        except jwt.ExpiredSignatureError:
+            return jsonify({
+                'message': 'Signature expired. Please log in again.'}), 401
+        except jwt.InvalidTokenError:
+            return jsonify({
+                'message': 'Invalid token. Please Provied a valid token!'
```
SannyTee: @JoyLubega there is a typo in the message being returned - "provied".

shemogumbe: + 1

helpers/auth/token_authentication.py
```diff
@@ -3,16 +3,14 @@
 
 
 class Authentication:
```
shemogumbe: When a module only has one class, it is advisable to give it a name that resembles the name of the main class in it

PR #41, opened 2018-06-13
#158326634 Enable backward relationship

PatrickCmd:
**What does this PR do?**

- Fix Bug on backward relationship

**Description of the task to be completed**

- For example Get all rooms should return a list of rooms, floor, block, location for particular room 
   object.

- N/A

- Executing the query below
allRooms {
id
name
capacity
floorId
}

- Response:
{
"data": {
"allRooms": [
{
"id": "3",
"name": "Gulu",
"capacity": 4,
"roomType": "Meeting Room",
"floorId": 1
},
{
"id": "4",
"name": "Entebbe",
"capacity": 8,
"roomType": "Meeting Room",
"floorId": 1
}
}
With this query it clearly shows that we only have floorId for a given room object which does not satisfy what the frontend mockup need. This means making another query to get more information needed which is not best practice with GraphQL

- With GraphQL the expected query would be
allRooms {
id
name
capacity
floorId
floor {
id
name
block {
id
name
location {
id
name
}
}
}
}

- #158326634

**Screenshots (if appropriate)**

screen shot 2018-06-13 at 5 40 25 pm

### DIFF
--- a/api/block/models.py
+++ b/api/block/models.py
@@ -11,4 +11,5 @@ class Block(Base, Utility):
     id = Column(Integer, primary_key=True)
     name = Column(String, nullable=False)
     location_id = Column(Integer, ForeignKey('locations.id'))
+    location = relationship('Location')
     floors = relationship('Floor')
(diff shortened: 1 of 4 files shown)


PatrickCmd:
@shemogumbe No they do not.

PR #48, opened 2018-06-22
[Bug #158140829] Restrict room resource addition to  admin users

dkam26:
#### What does this PR do?
- Restrict room resource addition to only authenticated admin users only

#### Description of the task to be completed**

- Room resources can currently be added by any users.Restrict addition of room resources to only authenticated admin users.

- #158140829

#### Screenshots

image


PatrickCmd:
@dkam26 Please correct the formatting of words on your PR.

tests/test_room_resource/test_create_room_resource.py
```diff
@@ -4,6 +4,10 @@
     resource_mutation_0_value_room_id_query, error_0_value_room_id,
     resource_mutation_empty_name_string_query, error_empty_name_string
 )
```
PatrickCmd: @dkam26 I think some of these imports are not required and used. So they should be removed and then flake8 will pass.

PatrickCmd:
@dkam26 Your PR is behind develop, please rebase it accordingly with develop.

tests/test_room_resource/test_create_room_resource.py
dkam26: @PatrickCmd ,Thanks for the feedback.I have made the necessary changes

PR #57, opened 2018-07-03
#158720631 Rename block model to office

SannyTee:
#### What does this PR do?
- Rename the block model to office 

- pull this branch `ch-rename-model-158720631`
- update database by running migration `Alembic upgrade head`
- The office model will reflect in the database with block model removed

n/a
[158720631](https://www.pivotaltracker.com/story/show/158720631)
n/a
#### Questions:
n/a


proxiex:
@shemogumbe the block folder has been deleted it is only showing up there because of the changes made to it. Git has to reflect it.

api/office/models.py
```diff
@@ -6,10 +6,9 @@
 from api.location.models import Location  # noqa: F401
 
 
-class Block(Base, Utility):
-    __tablename__ = 'blocks'
+class Office(Base, Utility):
+    __tablename__ = 'offices'
     id = Column(Integer, primary_key=True)
     name = Column(String, nullable=False)
     location_id = Column(Integer, ForeignKey('locations.id'))
```
PatrickCmd: Please include **location = relationship('Location')** for backward relationship for office with location.
Also rename the **block** package(folder) to **office**

proxiex: When that is done it create's some sort of loop.

PatrickCmd: I'm not following because we had this when it was still block.

SannyTee: @PatrickCmd The backward relationship between entities created an infinite sort of nested object.

PatrickCmd: @SannyTee Yes it is the practice with graphql. So doing query for office and you need a location for each, backward relationship does it and if you inspect well, you will find this implemented in each model.

SannyTee:
@shemogumbe  it's just as @proxiex mentioned. The block folder was renamed to office.

api/office/models.py
SannyTee: @PatrickCmd from a user point of view, if the backward relationship is included, the graphql docs will reflect the change and it can be confusing for someone trying to consume the API.

SannyTee: Thanks, @PatrickCmd. From the design specifications, I can see why the backward relationship is needed. The PR will be updated.

PR #58, opened 2018-07-03
#158581899 Alter all rooms query to to filter by location

dkam26:
#### What does this PR do?

   When the location parameter is provided, the query should  only return rooms in that location:
    else it should return all rooms

    #158581899

#### Screenshots 
   
     When a location parameter is provided.

screen shot 2018-07-03 at 12 09 16 pm

  When no location is provided.

screen shot 2018-07-03 at 1 24 33 pm


api/room/models.py
```diff
@@ -14,8 +15,10 @@ class Room(Base, Utility):
     capacity = Column(Integer, nullable=False)
     image_url = Column(String)
     calendar_id = Column(String)
+    location = Column(String, ForeignKey('locations.name'))
     floor_id = Column(Integer, ForeignKey('floors.id'))
     floor = relationship('Floor')
+    locations = relationship('Location')
```
PatrickCmd: change this field to singular **location** instead of **locations**

dkam26:
@proxiex ,all changes were made.

api/room/schema.py
```diff
@@ -45,6 +46,7 @@ class Arguments:
         capacity = graphene.Int()
         image_url = graphene.String()
         calendar_id = graphene.String()
+        location = graphene.String()
```
shemogumbe: Any reason why the location field was added, thought this was to be found by following back the foreign key relationship?

api/room/models.py
```diff
@@ -18,7 +18,7 @@ class Room(Base, Utility):
     location = Column(String, ForeignKey('locations.name'))
     floor_id = Column(Integer, ForeignKey('floors.id'))
     floor = relationship('Floor')
-    locations = relationship('Location')
+    location_name = relationship('Location')
```
shemogumbe: Any reason for specifically calling this 'location_name' and not location to reference the table Location?

PR #60, opened 2018-07-03
#158581852 country and timezone attributes

JoyLubega:
#### What does this PR do?
Add country and time_zone attributes in the location model as a choice field

#### Description
A user can add a location with country and its time_zone as selected fields. If the country is not part of the countries where operates, An error is returned.


api/location/models.py
```diff
@@ -10,5 +11,7 @@ class Location(Base, Utility):
     id = Column(Integer, primary_key=True)
     name = Column(String, nullable=False)
     abbreviation = Column(String, nullable=False)
+    country = Column(Enum('Uganda', 'Kenya', 'Nigeria', name='al_country'))
```
shemogumbe: How about we put the list on a different section using the python enums, checkout https://stackoverflow.com/questions/36136112/defining-sqlalchemy-enum-column-with-python-enum-raises-valueerror-not-a-valid -use this analogy for specifying timezones and locations on a different class and just pass them to the Enum column, as such it would improve the columns readability

api/location/models.py
```diff
@@ -10,5 +11,7 @@ class Location(Base, Utility):
...
+    time_zone = Column(Enum('GTM+3', 'GMT+3', 'GMT+1', name='all_timezones'))
```
shemogumbe: Same case as above

JoyLubega: Okay, let me fix it

JoyLubega:
@shemogumbe , i have fixed it

api/location/models.py
```diff
@@ -3,15 +3,28 @@
 
 from helpers.database import Base
 
+
 from utilities.utility import Utility
+import enum
+
+
+class Country_type(enum.Enum):
```
shemogumbe: In python, classes are never names with underscores, always use CountryType syntax -underscores are used in naming functions

api/location/models.py
```diff
@@ -3,15 +3,28 @@
...
+    Uganda = "Uganda"
+    Kenya = "Kenya"
+    Nigeria = "Nigeria"
+
+
+class TimeZone_type(enum.Enum):
```
shemogumbe: Always do not name classes using undersocores, use CapWords  instead refer https://www.python.org/dev/peps/pep-0008/#class-names

JoyLubega:
This file keeps changing..

utilities/utility.py
```diff
@@ -12,6 +12,36 @@ def validate_empty_fields(**kwargs):
...
+    Function to validate country fields when
+    saving an object
+    :params kwargs
+    """
+    my_args = []
+    timezones = ['EAST_AFRICA_TIME', 'WEST_AFRICA_TIME']
+    for value in kwargs:
+        time_zone_args = kwargs.get(value)
+        my_args.append(time_zone_args)
+    if my_args[3] not in timezones:
```
shemogumbe: You have an error in this validator, remember you only have 2 items in the list, and you looking for one with inex of 3 which is ideally the 4th item, this will always fail!

JoyLubega: let me check it out  now.

JoyLubega: @shemogumbe  it okay

JoyLubega: its not an error
screen shot 2018-07-10 at 18 46 54
Am picking form this list of arguments not the enumerators

shemogumbe: And btw, how many arguments should we pass this validator function, though just one would be enough, would it be necessary to pass multiple ?
I think we need to:
1. Pass just one argument.
2. it should even not be a keyword argument

PR #64, opened 2018-07-18
#159002990 Resolve code smell issues

mutoromike:
**What does this PR do?**
Fixes currently existing code smells.

**What are the relevant Pivotal Tracker stories?**
ID: #159002990
Story: Resolve code smell issues


mutoromike:
Mostly, the files in the helpers folder are files that interact with third-party applications, such as the Google Calendar and the Authentication, which as Converge we do not have major control over. That's why I was advised to ignore the folder.

PR #66, opened 2018-07-27
#159327828 Update Location Details

Bonifase:
#### What does this PR do?
 Allow users to have an end point from which they can update location details,
they can update info such as `imageURls`, `name`, `abbreviation` and `country` as well as `timezone`
- Run the server
- Run `updateLocation` mutation

- https://www.pivotaltracker.com/story/show/159327828


api/location/schema.py
```diff
@@ -30,6 +31,33 @@ def mutate(self, info, **kwargs):
...
+        location = Location.get_query(info)
+        location_object = location.filter(
+            LocationModel.id == location_id).first()
+        if not location_object:
+            raise GraphQLError("Location not found")
+        if "time_zone" in kwargs:
+            validate_timezone_field(**kwargs)
+        if "country" in kwargs:
+            validate_country_field(**kwargs)
+        if "name" in kwargs:
```
shemogumbe: You might need to check if all fields beside the fields already validated are available and validate if they are empty, this would only validate for the name field, remember this leaves, imageURL, abbreviation and timezone fields un validate

Bonifase: time_zone is already taken care of using `validate_timezone_field`, so I will consider validating empty abbreviation and image URL  together with the name.

PR #68, opened 2018-08-02
#159075759 paginate resource query 

mutoromike:
**What does this PR do?**
Allows pagination on the room resources
Allows resources to be returned without pagination

**Description of the task to be completed**
A user should specify page number and item per page

screen shot 2018-08-02 at 19 42 30

Allows all room resources to be returned if pagination's page and per page have not been provided

screen shot 2018-08-02 at 19 45 05

Create room resources with the required mutation
Provide pagination's page and perPage as shown in the image above (first image)
To test functionality without pagination, just run the query normally (see second image)

**What are related Pivotal Tracker stories?**
Story: Paginate resources query
Story ID: #159075759


api/room_resource/schema.py
```diff
@@ -14,6 +14,67 @@ class Meta:
         model = ResourceModel
 
 
+class PaginatedResource(graphene.ObjectType):
+    pages = graphene.Int(pages=graphene.Int())
```
shemogumbe: When we pass a value to a field, it should be an expression that when executed sets the default value for the attribute, or just the default value for the attribute e.g False for Boolean fields, It would then add  expressions such as `pages=graphene.Int()` for pages,  has_next and has_previous attributes, consider removing the values passed to this attributes, remember we set the values in the resolvers

api/room_resource/schema.py
```diff
@@ -14,6 +14,67 @@ class Meta:
...
+            return query.all()
+
+        if page:
+            if page < 1:
+                return GraphQLError("No page requested")
+
+            page = page - 1
+            self.query_total = query.count()
+            result = query.limit(per_page).offset(page*per_page)
+            # import pdb; pdb.set_trace()
```
SannyTee: @mutoromike It seems we are not making use of the commented line

mutoromike: Thanks, this has been fixed!

peterwade153:
LGTM

PR #70, opened 2018-08-03
chore(monitoring): Set up monitoring for application

kensanni:
#### What does this PR do?
 - This PR report errors from the API to `bugsnag` dashboard and also notify slack
- Import a module that doesn't exist
- Run the server

- #159326097
Notifitication on slack channel
screen shot 2018-08-06 at 3 32 07 am

Bugsnag Dashboard
screen shot 2018-08-03 at 12 09 15 pm

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -1,12 +1,21 @@
 import os
 
+import bugsnag
 from flask_script import Manager, Shell
+from bugsnag.flask import handle_exceptions
+
+# Configure bugnsag
+bugsnag.configure(
+  api_key=os.getenv('BUGSNAG_API_TOKEN'),
+  release_stage="development",
+  project_root="app"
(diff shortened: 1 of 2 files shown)


manage.py
```diff
@@ -1,12 +1,20 @@
 import os
+import bugsnag
 
 from flask_script import Manager, Shell
+from bugsnag.flask import handle_exceptions
```
shemogumbe: Since these are essentially bugsnag configurations, would you consider putting them in the Config.py file under ProductioConfig class

kensanni:
@shemogumbe, Putting the bugsnag configuration in the config file won't allow bugsnag to work. It has to be defined at the top level of the entry(app) file.

PR #71, opened 2018-08-06
#159461635 Admin can delete an admin

SannyTee:
#### What does this PR do?
 - This PR  enables  admin to delete an admin

- Run the server
- Run `deleteUser` mutation with userId that has admin privileges 

- #159461635
screen shot 2018-08-06 at 6 58 00 am


api/user/schema.py
```diff
@@ -52,15 +50,11 @@ def mutate(self, info, user_id, **kwargs):
         query_user = User.get_query(info)
         exact_query_user = query_user.filter(
             UserModel.id == user_id).first()
```
shemogumbe: Consider using emails to search for users, emails will be unique across Andela and better to use than id which might change from one Andela app to another

PR #84, opened 2018-08-16
#159652407 fix get rooms in block query

mutoromike:
**What does this PR do?**
Ensures that a list of rooms in a block is returned when the query `getRoomsInABlock` is executed. Previously, the block was returned instead, and not the desired output, rooms.

- Run the server
- Run getRoomsInABlock query

screen shot 2018-08-16 at 11 39 40

**What are relevant PivotalTracker stories?**
Story: Fix get rooms in block query
Id: #159652407


kevinbett:
Looks good Mike.

PR #85, opened 2018-08-16
#159817747 Add unique filter for resource instances

peterwade153:
### What does this PR do?
This PR enables querying for unique instances of resources

- Run the server

- Run getRoomsInALocation query

#159817747

When no value is passed
image

When a user set Unique to true
image


mutoromike:
Good work! Could you add the test-cases to assert the use of both `(unique:true)` and `(unique:false)`?

peterwade153:
I have added the test, for the `unique=false` is the same for all resources and it's already tested.

PR #94, opened 2018-09-03
#160235030 Delete office mutation

kevinbett:
### What does this PR do?

- This PR allows an admin to delete an office within his location.

- Run the server.
- Test the mutation at localhost:5000/mrm
- Run `deleteOffice` mutation with an `officeId` as the argument.

#160235030

###### deleting an office e.g the crest.
success

###### When not authorised to delete.
notauthorised

###### When office does not exist.
officenotfound

### DIFF
--- a/api/office/schema.py
+++ b/api/office/schema.py
@@ -32,6 +32,24 @@ class CreateOffice(graphene.Mutation):
         return CreateOffice(office=office)
 
 
+class DeleteOffice(graphene.Mutation):
+
+    class Arguments:
+        office_id = graphene.Int(required=True)
+    office = graphene.Field(Office)
+
+    @Auth.user_roles('Admin')
+    def mutate(self, info, office_id, **kwargs):
+        query_office = Office.get_query(info)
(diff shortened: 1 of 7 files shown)


helpers/room_filter/room_filter.py
```diff
@@ -56,6 +56,11 @@ def location_join_resources():
     return location_query
 
 
+def location_join_office():
+    location_query = Location.query.join(Office)
+    return location_query
```
mutoromike: Hello @kevinbett , I don't think we need this because we are not doing any joins here. In the `admin_roles.py` file just import the `Location Model` and we will be good to go.

mutoromike:
It will also be nice to add a test case for the `delete office` functionality

helpers/room_filter/room_filter.py
kevinbett: Hi @mutoromike I have made the changes requested above.

kevinbett:
@mutoromike I have added the `delete office` test cases.

mutoromike:
Great! 👍

PR #95, opened 2018-09-04
#160193993 Add error handling for Roles and Users

sam-karis:
#### What does this PR do?
This PR handle conflict/duplication error for already existing Roles and Users

- Run the server.    
- Test the queries at `localhost:5000/mrm`
- Run createUser mutation with an existing user email
- Run createRole mutation with an existing role name

[Error handling](https://www.pivotaltracker.com/n/projects/2154921/stories/160193993)

##### createUser error snapshot
image

##### createRole error snapshot
image

### DIFF
--- a/api/role/schema.py
+++ b/api/role/schema.py
@@ -2,6 +2,7 @@ import graphene
 
 from graphene_sqlalchemy import (SQLAlchemyObjectType)
 from api.role.models import Role as RoleModel
+from helpers.auth.error_handler import SaveContextManager
 
 
 class Role(SQLAlchemyObjectType):
@@ -18,9 +19,8 @@ class CreateRole(graphene.Mutation):
 
     def mutate(self, info, **kwargs):
         role = RoleModel(**kwargs)
-        role.save()
-
-        return CreateRole(role=role)
+        with SaveContextManager(role, kwargs.get('role'), 'Role'):
+            return CreateRole(role=role)
 
 
 class Query(graphene.ObjectType):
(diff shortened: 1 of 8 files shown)


mutoromike:
I think you can remove the `Conflict error!` bit and just have the error itself returned

peterwade153:
Good work @sam-karis , just do away with `Conflict error!` and your good to go.

kevinbett:
LGTM once changes mentioned above are implemented.

sam-karis:
@mutoromike, @peterwade153, @kevinbett I have removed the `Conflict error!` bit on the error message.

Bonifase:
LGTM

api/role/schema.py
```diff
@@ -17,6 +18,9 @@ class Arguments:
     role = graphene.Field(Role)
 
     def mutate(self, info, **kwargs):
+        new_role = RoleModel.query.filter_by(role=kwargs.get('role')).first()
```
shemogumbe: Use sqlalchemy exceptions to catch more errors, e.g db conflict, db connection, validation errors etc, this can be in a try except block or wrapped in a context manager, using a context manger will help us solve for duplication as well

api/user/schema.py
```diff
@@ -26,6 +27,10 @@ class Arguments:
     user = graphene.Field(User)
 
     def mutate(self, info, **kwargs):
+        new_user = UserModel.query.filter_by(
+            email=kwargs.get('email')).first()
```
shemogumbe: Same comment as above

sam-karis:
@shemogumbe I have implemented the changes requested to catch all SQLAlchemyError

PR #96, opened 2018-09-04
#160193943 Add error handling for Rooms and Offices

kzyGit:
#### What does this PR do?
- Handles duplication errors on creation of rooms and offices with an already existing name.

- Run the server
- Run `createOffice` mutation using an existing office name
- Run `createRoom` mutation using an existing room name

[#160193943](https://www.pivotaltracker.com/story/show/160193943)

#### Screenshots
- createOffice mutation
screen shot 2018-09-05 at 15 51 14

- createRoom mutation
screen shot 2018-09-05 at 15 49 19


peterwade153:
Hey Kezzy Good work there, could we have a message like `St.catherine's office already exists ` rather than adding `conflict error` because this is what the frontend will pick.

mutoromike:
Good work with test cases and helper to handle the issue.

kzyGit:
Well noted @peterwade153 . Fixing that in a bit

kevinbett:
Good work @kzyGit You will be good to go after implementing the recommendations mentioned above.

api/office/schema.py
```diff
@@ -22,6 +23,10 @@ class Arguments:
 
     @Auth.user_roles('Admin')
     def mutate(self, info, **kwargs):
+        existing_office = OfficeModel.query.filter_by(
```
shemogumbe: This would only check for one error, the conflict, it would be great to use SQLALCHEMY EXCEPTIONS and overide the messages, as such we will catch lots of errors includuing db connection, conflicts, invalid request, invalid arguments etc, this is better used in a try except block.
we have 2 options:
-Use a try except block in the mutation
-Build a context manager which would encapsulate the try except block with sqlalchemy exceptions

helpers/auth/validator.py
```diff
@@ -17,4 +18,11 @@ def assert_wing_is_required(office, kwargs):
 
 def verify_email(email):
     return bool(re.match('^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+$',
-                email))
+                         email))
+
+
+class ErrorHandler():
+
+    def check_conflict(self, entity, entity_name):
```
shemogumbe: Looks good but only fixes one error, check request above for how to handle more errors.

api/office/schema.py
kzyGit: Well noted @shemogumbe . Working on it.

kzyGit:
@shemogumbe I have implemented the above feedback and used a try except block in the mutation.
@peterwade153 Feedback on updating response and room model implemented

helpers/room_filter/room_filter.py
```diff
@@ -95,5 +110,8 @@ def room_filter(query, filter_data):
         query = query.filter(RoomModel.capacity == capacity)
         query = query.filter(Resource.name.ilike('%' + resources + '%'))
         return query.filter(Location.name.ilike('%' + location + '%'))
+    # elif office:
+    #     query = room_join_office(query)
+    #     return query.filter(Office.id.ilike('%' + office + '%'))
```
peterwade153: Lets have this removed, if it's necessary

kzyGit:
@kevinbett already resolved the conflicts. Thank you.

PR #98, opened 2018-09-06
#160235195 Update Create Room Mutation to include Block_id and Wing_id

Bonifase:
#### What does this PR do?
 Enable `createRoom` mutation for Nairobi to include `blockId`, and `createRoom` mutation for Lagos to include `wingId` when admin creates a meeting room
- Clone this branch in your local machine, Run the server
- Run `createRoom` mutation with `blockId` when the office is Nairobi or `wingId` when the office is Lagos in the keyword arguments
https://www.pivotaltracker.com/story/show/160127115
#### Screenshots 
     When an admin runs the mutation with block id.
screen shot 2018-09-06 at 11 07 44
     When an admin runs the mutation with wing id.
screen shot 2018-09-06 at 12 22 43

### DIFF
--- a/api/room/schema.py
+++ b/api/room/schema.py
@@ -5,6 +5,8 @@ from graphql import GraphQLError
 
 from api.room.models import Room as RoomModel
 from api.office.models import Office
+from api.block.models import Block
+from api.floor.models import Floor
 from helpers.calendar.events import RoomSchedules
 from utilities.utility import validate_empty_fields, update_entity_fields
 from helpers.auth.authentication import Auth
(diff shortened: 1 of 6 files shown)


api/room/schema.py
```diff
@@ -49,13 +52,33 @@ def mutate(self, info, office_id, **kwargs):
...
         if not get_office:
             raise GraphQLError("No Office Found")
-
+        if 'block_id' in kwargs:
+            exact_block = Block.query.filter_by(id=kwargs['block_id']).first()
+            if not exact_block:
+                raise GraphQLError("Block with such id does not exist")
+            floor = Floor.query.filter_by(
+                block_id=exact_block.office_id).first()
+            if floor.id == kwargs['floor_id']:
```
peterwade153: A Block can have multiple floors, but from this query(line 59) we only return the first floor instead of a list of floors. Therefore if the `floor_id` for an existing floor is passed, but doesn’t match the one returned from the Floor query, the room won’t be registered.

Bonifase: Yes @peterwade153, the room should be created in the floor whose `floor id` is passed, so we query if that floor exists in that particular block whose `block id` is also passed then we create the room, otherwise, we don't create the room having in mind that `floor id` is unique for each floor.

mutoromike: I agree a room is only created if the block and floor both exist, 👍 @Bonifase

kzyGit:
Good job @Bonifase . You however have merge conflicts that you need to resolve first.

PR #103, opened 2018-09-18
#160445879 Analytics for the least used room per week

sam-karis:
#### What does this PR do?
This PR handle analytics for the least used room per week

- Run the server.    
- Test the queries at `localhost:5000/mrm`
- Run `analyticsForRoomLeastUsedPerWeek` query with following params.  
    - `locationId`
    - `weekStart`
    - `weekEnd`

[Analytics for the least used room per week](https://www.pivotaltracker.com/n/projects/2154921/stories/160445879)

##### Snapshot of a period where all rooms have events booked
image

##### Snapshot of a period where some rooms have no events booked
image

### DIFF
--- a/tests/test_rooms/test_room_analytics.py
+++ b/tests/test_rooms/test_room_analytics.py
@@ -17,32 +17,34 @@ from fixtures.room.room_analytics_fixtures import (
 
 class QueryRoomsAnalytics(BaseTestCase):
 
-    api_headers = {"Authorization": "Bearer" + " " + admin_api_token}
-
     def test_most_used_room_in_a_month_analytics(self):
+        headers = {"Authorization": "Bearer" + " " + admin_api_token}
         response = self.app_test.post(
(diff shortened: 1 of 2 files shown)


Bonifase:
Great job! I would suggest that you also handle analytics for rooms in other locations other than only with `locationId=1`, we have three locations, that is Nairobi, Kampala, and Lagos.

tests/test_rooms/test_room_analytics.py
```diff
@@ -0,0 +1,30 @@
...
+    get_least_used_room_per_week_response,
+    get_least_used_room_without_event_query,
+    get_least_used_room_without_event_response
+
+)
+
+
+class QueryRoomsAnalytics(BaseTestCase):
+
+    api_headers = {"Authorization": "Bearer" + " " + admin_api_token}
```
shemogumbe: Since we changed all headers to use the variable 'headers' and noy 'api_header' all across the application, it would make since to just use a uniform name  for consitency even as we make plans to collect them and reuse them instead of declaring them in each unit test to avoid duplication,

PR #113, opened 2018-10-01
#160845129 Add cascade effect on  delete office

sam-karis:
### What does this PR do?

- This PR adds the cascade effect on delete office to ensure all data related to an office is deleted when an office is deleted.

- Run the server.
- Test the query at localhost:5000/mrm.
- Run `deleteOffice` mutation with `officeId` as an argument.

#160845129

###### Deleting an office with one block
image

###### Deleting an office with more than one block
image

### DIFF
--- a/api/block/models.py
+++ b/api/block/models.py
@@ -12,4 +12,4 @@ class Block(Base, Utility):
     name = Column(String, nullable=False)
     office_id = Column(Integer, ForeignKey('offices.id'))
     offices = relationship('Office')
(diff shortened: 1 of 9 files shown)


sam-karis:
@Bonifase, No the changes do not cause any change in the database only introduce the cascade effect.

Bonifase:
> @Bonifase, No the changes do not cause any change in the database only introduce the cascade effect.

Cool

PR #143, opened 2018-11-05
#161676333 fix create user role duplication

migot01:
**What does this PR do?**

It ensures that a user role can only be created once, and if one tried to create the same user role again, then it returns a descriptive error message

- Run the server.
-Test the queries at `localhost:5000/mrm`
-Run the following query: 
`createUsersRole`
It should be successful when you run it once but should fail on the second run

What are the relevant pivotal tracker stories?
[#161676333](https://www.pivotaltracker.com/story/show/161676333)

##### Screenshots for first time run
image
##### Screenshots for second time run
image

### DIFF
--- a/api/user_role/schema.py
+++ b/api/user_role/schema.py
@@ -1,4 +1,5 @@
 import graphene
+from graphql import GraphQLError
 
 from graphene_sqlalchemy import (SQLAlchemyObjectType)
 
@@ -19,6 +20,11 @@ class CreateUserRole(graphene.Mutation):
     user_role = graphene.Field(UsersRole)
 
     def mutate(self, info, **kwargs):
+        user_role = UserRoleModel.query.filter_by(user_id=kwargs['user_id'],
+                                                  role_id=kwargs['role_id']).\
+                                                  all()
+        if user_role:
+            raise GraphQLError("You cannot create user role twice")
         user_role = UserRoleModel(**kwargs)
         user_role.save()
 


api/user_role/schema.py
```diff
@@ -19,6 +20,10 @@ class Arguments:
     user_role = graphene.Field(UsersRole)
 
     def mutate(self, info, **kwargs):
+        roles = UserRoleModel.query.filter_by(user_id=kwargs['user_id']).\
```
sam-karis: @migot01 I would suggest you collapse the two `filter_by()` into one and also in the if condition I think `len` is not necessary  as simple `if roles:` is better.

api/user_role/schema.py
```diff
@@ -19,6 +20,10 @@ class Arguments:
     user_role = graphene.Field(UsersRole)
 
     def mutate(self, info, **kwargs):
+        roles = UserRoleModel.query.filter_by(user_id=kwargs['user_id'],
```
shemogumbe: Hello, this query checks if a users already has roles and throws errors if we add more roles right, which should not be the case coz a user can ideally have multiple roles, e.g can be ops associate and admin, so if one had been added, another role can still be added, thought it was to prevent having same role twice for one user

migot01: hey @shemogumbe, actually it does not throw an error if we add more roles to the user, a user can have multiple roles,  it only throws an error if we try to create the same user with the same role more than once ie user_1 with role_1 can only be created once but we can create  user_1 with role_2 which should be the case

migot01: currently, we can create use_1 with role_1 severally and it duplicates itself on the database

PR #145, opened 2018-11-09
#161778500 Ratio of cancelled meetings to bookings

3V0L:
#### What does this PR do?
- Adds the functionality to get the ratio of bookings to cancelled meetings and check ins for an individual room. It also adds on a percentage of cancellations and checkins

- Run the server 
- On `localhost:5000/mrm` run the `analyticsRatios` query with `startDate` and `endDate` as parameters
- This should return the total ratio for all the rooms put together. A screenshot example will be attached below
- On the same URL run the query `analyticsRatiosPerDay` with `startDate` and `endDate` as parameters
- This should return the bookings to cancellation/checkins ratio of each individual room. A screenshot example will be attached below

[#161778500](https://www.pivotaltracker.com/story/show/161778500)

#### Screenshots 
**Query for sum of all room ratios**
screenshot 2018-11-09 at 16 06 56

**Query for individual room ratios**
screenshot 2018-11-12 at 11 12 14


api/analytics/analytic_report.py
```diff
@@ -54,7 +54,6 @@ def get_least_used_rooms(self, all_rooms_summary_df):
         '''
         Get analytics for the 10 least used rooms
         '''
-        # import pdb; pdb.set_trace()
```
shemogumbe: Remove commented out code from the source control

3V0L: This comment was merged in a previous PR, I was removing it from the source code

PR #148, opened 2018-11-21
#162054093 Enable admin create blocks in Nairobi office

migot01:
#### What does this PR do?
- Adds create block mutation to enable an admin to create blocks in the Nairobi office.

- Run the server
- Run `createBlock` mutation with the parameter 'room_id'  and 'name', as  required parameters

[#162054093](https://www.pivotaltracker.com/story/show/162054093)

#### Screenshots
1. Create block
image

2. Try duplicate same block name
image

3.Create block without name
image

4. Try to create a block that is not in the Nairobi office
image

5. Error message for the non-existent office
image

### DIFF
--- a/api/block/schema.py
+++ b/api/block/schema.py
@@ -24,11 +24,11 @@ class CreateBlock(graphene.Mutation):
     @Auth.user_roles('Admin')
     def mutate(self, info, **kwargs):
         validate_empty_fields(**kwargs)
-        block_name = BlockModel.query.filter_by(name=kwargs['name']).all()
-        if len(block_name) > 0:
+        block_name = BlockModel.query.filter_by(name=kwargs['name'].title()).all() # noqa
+        if block_name:
             raise GraphQLError("Block aleady exists")
         get_office = Office.query.filter_by(id=kwargs['office_id']).first()
-        if get_office is None:
+        if not get_office:
             raise GraphQLError("Office not found")
         location = get_office.location.name
         if location.lower() == 'nairobi':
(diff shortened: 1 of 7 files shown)


migot01:
@3V0L I have made the requested changes please have another look at it

api/block/schema.py
```diff
@@ -1,16 +1,43 @@
...
+    class Arguments:
+        name = graphene.String(required=True)
+        office_id = graphene.Int(required=True)
+    block = graphene.Field(Block)
+
+    @Auth.user_roles('Admin')
+    def mutate(self, info, **kwargs):
+        validate_empty_fields(**kwargs)
+        block_name = BlockModel.query.filter_by(name=kwargs['name'].title()).all() # noqa
+        if len(block_name) > 0:
```
shemogumbe: In stead of checking the length of the list after querying the blocks model with the name, you can simply use truthy test, something like if if block_name(this is called implicit boleanness and is more pythonic code)

api/block/schema.py
```diff
@@ -1,16 +1,43 @@
...
+            raise GraphQLError("Block aleady exists")
+        get_office = Office.query.filter_by(id=kwargs['office_id']).first()
+        if get_office is None:
```
shemogumbe: simply use implicit boleannes here, like if get_office: ...

migot01: @shemogumbe  cool on it

migot01: @shemogumbe  I have impemented the changes

PR #155, opened 2018-12-05
#162233289 Enable admin create feedback question and users respond to the questions

migot01:
#### What does this PR do?
- Adds questions model and mutation to enable admin to create feedback questions.
- Add user response models and mutations to enable a user to rate, check if all equipment is available in a room or not and add any other suggestions

- First,  Run migrations to add Question and Response model. 
- Run the server
#### Admin can create feedback question
image
#### error while creating a question with a missing field
image

#### error while creating a question without admin rights
image
#### User able to rate a room 
image
#### Trying to rate a room that does not exist
image
#### Trying to rate on a question that is not about rating
image
#### Rating a question that does not exist
image
#### Trying to rate with a value that is not between 1 and 5
image
#### Checking if a room has all tools or not this can only be a boolean value
image
#### A user able to add any other suggestions
image


api/response/models.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class Response(Base, Utility):
+    __tablename__ = 'responses'
+    id = Column(Integer, primary_key=True)
+    room_id = Column(Integer, ForeignKey('rooms.id'))
+    question_id = Column(Integer, ForeignKey('questions.id'))
+    rate = Column(Integer, nullable=True)
+    check = Column(Boolean, nullable=True)
+    text_area = Column(String, nullable=True)
```
lennykioko: What do you think about having a field to timestamp the time at which a particular response was given?

PR #158, opened 2018-12-06
#162236838: Admin should be able to update and delete feedback question

cwizard2011:
**What does this PR do?**
- Enables an admin to delete and update feedback questions

- Run the `deleteQuestion` mutation with the `questionId`
- Run the `updateQuestion` mutation with the `questionId` and any other field to be updated

[#162236838](https://www.pivotaltracker.com/story/show/162236838)

**Screenshots**
- Delete question
screen shot 2018-12-06 at 1 02 15 pm

- Update question
screen shot 2018-12-06 at 1 02 28 pm


tox.ini
```diff
@@ -37,4 +37,4 @@ exclude =
     alembic,
     .circleci
 max-complexity = 10
-max-line-length = 80
+max-line-length = 80
```
migot01: Add an end of line, otherwise nice work

cwizard2011: Thanks for the feedback, I have implemented the feedback

api/question/schema.py
```diff
@@ -0,0 +1,73 @@
...
+        question = graphene.String()
+        start_date = graphene.String()
+        end_date = graphene.String()
+
+    question = graphene.Field(Question)
+
+    @Auth.user_roles('Admin')
+    def mutate(self, info, question_id, **kwargs):
+        validate_empty_fields(**kwargs)
+        query_question = Question.get_query(info)
```
shemogumbe: LGTM

api/response/models.py
```diff
@@ -0,0 +1,14 @@
...
+from utilities.utility import Utility
+
+
+class Response(Base, Utility):
+    __tablename__ = 'responses'
+    id = Column(Integer, primary_key=True)
+    user_id = Column(Integer, ForeignKey('users.id'))
+    room_id = Column(Integer, ForeignKey('rooms.id'))
+    question_id = Column(Integer, ForeignKey('questions.id'))
+    rate = Column(Integer, nullable=True)
```
shemogumbe: As it is, one question can have a text, rate and check answer,
I feel it would have been better if the type of answer depended on the type of question , so questions have a type and the anser would depend on the question.type file, and the logic can be implemented in the mutate method

api/response/schema.py
```diff
@@ -0,0 +1,100 @@
...
+from api.question.models import Question
+from helpers.auth.authentication import Auth
+
+
+class Response(SQLAlchemyObjectType):
+    class Meta:
+        model = ResponseModel
+
+
+class CreateRate(graphene.Mutation):
```
shemogumbe: This would be unnecessary if we had if had a question type and then in the response, restrict answers based on question.type

api/response/schema.py
```diff
@@ -0,0 +1,100 @@
...
+            validate_rating_field(**kwargs)
+        else:
+            raise GraphQLError("Select a rating question")
+        user = get_user_from_db()
+        rating = ResponseModel(**kwargs, user_id=user.id)
+        rating.save()
+        return CreateRate(rating=rating)
+
+
+class CreateCheck(graphene.Mutation):
```
shemogumbe: This would be unnecessary if we had if had a question type and then in the response, restrict answers based on question.type

api/response/schema.py
```diff
@@ -0,0 +1,100 @@
...
+        if question_type.lower() == 'check':
+            user = get_user_from_db()
+            checking = ResponseModel(**kwargs, user_id=user.id)
+            checking.save()
+        else:
+            raise GraphQLError("Select a check question")
+        return CreateCheck(checking=checking)
+
+
+class CreateSuggestion(graphene.Mutation):
```
shemogumbe: This would be unnecessary if we had if had a question type and then in the response, restrict answers based on question.type

api/response/models.py
cwizard2011: I didn't work on this file, I only rebased from @migot01 branch for creating feedback question

cwizard2011: I will have a sync with him to rectify this

api/response/schema.py
cwizard2011: I didn't create this, I will have a sync with @migot01, he created the mutation

api/response/schema.py
cwizard2011: I didn't create this, I will have a sync with @migot01, he created the mutation

api/response/schema.py
cwizard2011: I didn't create this, I will have a sync with @migot01, he created the mutation

api/response/models.py
cwizard2011: @shemogumbe I have implemented the requested changes

api/response/schema.py
cwizard2011: @shemogumbe I have implemented the requested changes

api/response/schema.py
cwizard2011: @shemogumbe I have implemented the requested changes

api/response/schema.py
cwizard2011: @shemogumbe I have implemented the requested changes

PR #161, opened 2018-12-10
#162240864: Admin can view a response summaries of room feedback question

omobosteven:
**What does this PR do?**
- Enables an admin to view the response summaries of feedback questions

- Run the `feedbackQuestion` query 

[#162240864](https://www.pivotaltracker.com/story/show/162240864)

**Screenshots**
- Cleanliness rating summary
screen shot 2018-12-10 at 3 23 10 pm

### DIFF
--- a/alembic/versions/a7945f07140d_add_room_cancellation_duration_column.py
+++ b/alembic/versions/a7945f07140d_add_room_cancellation_duration_column.py
@@ -11,7 +11,7 @@ import sqlalchemy as sa
 
 # revision identifiers, used by Alembic.
 revision = 'a7945f07140d'
-down_revision = '6046c7c7c022'
+down_revision = '814b5dc14d80'
 branch_labels = None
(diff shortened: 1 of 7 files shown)


fixtures/questions/get_question_fixtures.py
```diff
@@ -0,0 +1,26 @@
...
+      responseCount,
+      cleanlinessRating
+    }
+  }
+}
+'''
+
+get_question_query_response = {
+  "data": {
+    "feedbackQuestion": {
```
lennykioko: I am of the idea that we could get rid of this additional nesting created at the feedbackQuestion so that we can directly move from data to totalResponses and then responses, what do you think?

omobosteven: That is a good idea. I will make the changes.

kampkelly:
Looks good to me

PR #162, opened 2018-12-10
#162545107 Remove all tokens from the source code and add them to the environment variables.

lennykioko:
**What does this PR do?**

- Removes all tokens from the source code.
- Adds the tokens to the environment variables.
- Updates the `ADMIN_TOKEN`, `USER_TOKEN` and `ADMIN_NIGERIA_TOKEN` since the ones that we had have been revoked.

- Run the tests using `tox`.
- All the tests should pass just as they used to before.

What are the relevant pivotal tracker stories?
[#162545107](https://www.pivotaltracker.com/story/show/162545107)


tests/test_rooms/test_create_room.py
```diff
@@ -25,7 +24,7 @@ def test_room_creation(self):
         """
         Testing for room creation
         """
-        headers = {"Authorization": "Bearer" + " " + admin_api_token}
+        headers = {"Authorization": "Bearer" + " " + os.getenv('ADMIN_TOKEN')}
```
shemogumbe: Would it be nice to keep the token fixtures file, and have variables such ADMIN_TOKEN = os.getenv('ADMIN_TOKEN') instead of getting it from env file in all places.

lennykioko: I had done that at first, I then found that the instances that we used the tokens were few and thought that the current approach was easier. I will update this right away.

PR #170, opened 2018-12-20
#162725432 Admin activate or deactivate question

michael-basweti:
**What does this PR do?**
- Admin is able to activate and deactivate a question

- Run the following query
    ```
   mutation {
      updateQuestion(questionId:1,
      isActive: false) {
        question {
          id,
          isActive
        }
      }
    }
   ```

[#162240864](https://www.pivotaltracker.com/n/projects/2154921/stories/162725432)

screenshot 2018-12-19 at 12 08 57
screenshot 2018-12-19 at 12 09 25


api/question/models.py
```diff
@@ -13,3 +13,4 @@ class Question(Base, Utility):
     end_date = Column(String, nullable=False)
     total_views = Column(Integer, default=0)
     response = relationship('Response', cascade="all, delete-orphan")
+    is_active = Column(Boolean, nullable=True)
```
caveinn: I think for this it would be better if you set a default value to false rather than setting nullable to true

PR #171, opened 2018-12-20
#162727868 Enable admin add time to a question being created

RonKbS:
**What does this PR do?**
- Allows an admin to add a startTime and endTime to a question

- Run the following query

```mutation {
  createQuestion(questionType:"Rate",
  question:"How will you rate the brightness of the room",
  startDate:"2018-12-20T23:42:43", endDate:"2018-12-28T15:42:43") {
    question{
      id
      question
      questionType
      startDate
      endDate
      }
  }
}
```

[#162727868](https://www.pivotaltracker.com/story/show/162727868)

**Screenshots**
screen shot 2018-12-20 at 12 54 28
screen shot 2018-12-20 at 12 54 58

### DIFF
--- a/fixtures/questions/create_questions_fixtures.py
+++ b/fixtures/questions/create_questions_fixtures.py
@@ -1,18 +1,42 @@
+import datetime
+
+start_date = (
+  datetime.datetime.now().replace(microsecond=0) + datetime.timedelta(days=1)
+).isoformat()
+end_date = (
+  datetime.datetime.now().replace(microsecond=0) + datetime.timedelta(days=2)
+).isoformat()
+wrong_start_date = (
+  datetime.datetime.now().replace(microsecond=0) + datetime.timedelta(days=-1)
+).isoformat()
+wrong_end_date = (
+  datetime.datetime.now().replace(microsecond=0) + datetime.timedelta(days=1)
+).isoformat()
+
+
 create_question_query = '''
- mutation{
+ mutation{{
   createQuestion(questionType:"Rate",
   question:"How will you rate the brightness of the room",
-  startDate:"20 Nov 2018", endDate:"28 Nov 2018") {
-    question{
+  startDate:"{}", endDate:"{}") {{
+    question{{
       id
       question
       questionType
       startDate
       endDate
-      }
-  }
-}
-'''
+      }}
+  }}
+}}
+'''.format(start_date, end_date)
+
+create_question_query_with_early_startDate = create_question_query.replace(
+  start_date, wrong_start_date
+)
+
+create_question_query_with_early_endDate = create_question_query.replace(
+  end_date, wrong_end_date
+)
 
 create_question_response = {
   "data": {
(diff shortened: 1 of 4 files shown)


RonKbS:
@Bonifase the failing tests have been fixed

RonKbS:
Thank you for the feedback @joshuaocero.
I am going to work on the nested if-statements and the flagged variable/function names.
As for handling the responses in the calling function, logic concerns will be introduced into the pristine code below. I am of the view to, therefore, leave the responses in the utility file as has been done for the other functions in the same file.
screen shot 2018-12-21 at 11 48 38

blac-siren:
@RonKbS convert the start date and end date to pick standard time format i.e `_dd-mmm-yyyy hh:mm_`  instead of i.e `2018-12-20T23:42:43`

joshuaocero:
> Thank you for the feedback @joshuaocero.
> I am going to work on the nested if-statements and the flagged variable/function names.
> Regarding handling the responses in the calling function, logic concerns will be introduced into the pristine code below. I am of the view to, therefore, leave the responses in the utility file as has been done for the other functions in the same file.
> screen shot 2018-12-21 at 11 48 38

Okay @RonKbS... 
Please take note of my recommended function definition. It will help you remove unnecessary code and do away with hard-coded values. There is no need to check for `start_date` and `end_date` in kwargs as the function `validate_empty_fields` already that.

RonKbS:
@joshuaocero the check for start_date and end_date is used by the update mutation which may not have the date fields if it is only required to update the question only. The validate_date_range function will fail if the check was not included and someone tries to update the questions without the date fields.

RonKbS:
@blac-siren I missed your request to change the format hence my answering it late.
Looking at these sources...
 - https://www.cl.cam.ac.uk/~mgk25/iso-time.html
 - https://en.wikipedia.org/wiki/ISO_8601
It follows that the actual international format is the one implemented in this PR.
the 'T' in the mutation is required by the DateTime graphene fields but is not saved as can be seen in response shown in the screenshot of the PR.

fixtures/questions/create_questions_fixtures.py
```diff
@@ -1,18 +1,42 @@
...
+
+
 create_question_query = '''
- mutation{
+ mutation{{
   createQuestion(questionType:"Rate",
   question:"How will you rate the brightness of the room",
-  startDate:"20 Nov 2018", endDate:"28 Nov 2018") {
-    question{
+  startDate:"{}", endDate:"{}") {{
```
Bonifase: Since this is just a test fixture, I recommend that you just type the values the way they are supposed to be. e.g startDate: 2018-12-23T23:42:43. Avoid using string format unless it is very necessary. Also have a sync with the front end team to agree on the date format they will be sending to the backend api in their request.

RonKbS: I have consulted with a frontend team member and, going through an implemented date feature, we have been able to come up with the DateTime format used in this PR.
With regards to typing in the date, the implementation of the feature is such that one can not set a start date before the current date and an end date before a specified period of days. The dates entered need to be dynamic for the tests to pass.

Bonifase: Okay.

PR #176, opened 2018-12-21
#162572539 Enable admin to view tool availability and additional feedback

bevkololi:
**What does this PR do?**
- Admin is able to view the total tool availability of a room (missing tools) and the additional feedback given per room (comments given per room).

**Description of task to be completed**
- As an admin, I want to view the details of tool availability and additional comments feedback question, so that I can be able to see the response summaries from all the rooms on which these two questions were asked.

The data returned is response data in all rooms. The aggregate data e.g the number of missing items will be done in the front-end.

- Add questions using the questions mutation
- Provide feedback to any rooms available i.e rating, report any missing items, and give any additional feedback.

Run the following query:
```query {
    allRoomResponses {
	responses {
        roomName,
        totalResponses,
        response {
          responseId,
          rating,
          createdDate,
          suggestion,
          missingItems
        }
      }
    }
}
```
**What are the relevant pivotal stories?**
[#162241130](https://www.pivotaltracker.com/story/show/162572539)

**Screenshots**
This is what to expect:

screenshot 2019-01-15 at 18 31 16

### DIFF
--- a/api/response/schema_query.py
+++ b/api/response/schema_query.py
@@ -1,8 +1,9 @@
 import graphene
 from graphql import GraphQLError
-
 from api.response.schema import Response
+from api.response.models import Response as ResponseModel
 from api.room.schema import Room
+from api.room.models import Room as RoomModel
 from helpers.auth.authentication import Auth
 
 
@@ -20,9 +21,15 @@ class RoomResponse(graphene.ObjectType):
     response = graphene.List(ResponseDetails)
 
 
+class AllResponses(graphene.ObjectType):
+    room_name = graphene.String()
+    responses = graphene.List(RoomResponse)
+
+
 class Query(graphene.ObjectType):
     room_response = graphene.Field(
         RoomResponse, room_id=graphene.Int())
+    all_room_responses = graphene.Field(AllResponses)
 
     def get_room_response(self, test_response):
         response = []
@@ -70,3 +77,20 @@ class Query(graphene.ObjectType):
                 room_name=room_name,
                 total_responses=total_response,
                 response=responses)
+
+    @Auth.user_roles('Admin')
+    def resolve_all_room_responses(self, info):
+        response = []
+        rooms = RoomModel.query.all()
+        for room in rooms:
+            room_name = room.name
+            total_response = ResponseModel.query.filter_by(
+                room_id=room.id).count()
+            room_response = Response.get_query(info).filter_by(room_id=room.id)
+            all_responses = Query.get_room_response(self, room_response)
+            responses = RoomResponse(
+                room_name=room_name,
+                total_responses=total_response,
+                response=all_responses)
+            response.append(responses)
+        return AllResponses(responses=response, room_name=room_name)
--- a/fixtures/response/room_response_fixture.py
+++ b/fixtures/response/room_response_fixture.py
@@ -45,3 +45,46 @@ get_room_response_non_existence_room_id = '''{
     }
 }
 '''
+
+summary_room_response_query = '''{
+    allRoomResponses {
+        responses {
+        roomName,
+        totalResponses,
+        response {
+          responseId,
+          rating,
+          suggestion,
+          missingItems
+        }
+      }
+    }
+}
+'''
+
+summary_room_response_data = {
+  "data": {
+    "allRoomResponses": {
+      "responses": [
+        {
+          "roomName": "Entebbe",
+          "totalResponses": 2,
+          "response": [
+            {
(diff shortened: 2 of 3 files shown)


api/question/schema_query.py
```diff
@@ -32,9 +37,20 @@ def get_room_response(self, unique_responses, question_id, question_type):
...
-                average_rating = sum([response.rate for response in responses_in_room]) / responses_in_room.count()  # noqa: E501
-                room_response = ResponsePerRoom(room_name=room_name, response_count=response_count, cleanliness_rating=average_rating)  # noqa: E501
+            average_rating = sum([response.rate for response in responses_in_room]) / responses_in_room.count()  # noqa: E501
+            all_resources = ResourceModel.query.filter_by(room_id=room.room_id).count()  # noqa: E501
+            all_missing_resources = MissingItemsModel.query.filter_by(room_id=room.room_id).count()  # noqa: E501
+            total_missing_resources = '{} out of {}'.format(all_missing_resources, all_resources)  # noqa: E501
+            try:
+                last_additional_comment = ResponseModel.query.filter_by(room_id=room.room_id).filter(column("additional_comment").isnot(None))[-1].additional_comment   # noqa: E501
+            except Exception:
+                last_additional_comment = "No comments yet"
```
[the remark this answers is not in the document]
bevkololi: @Bonifase, the merge conflicts have been solved. The return/raise statement however did not work as we had discussed in the sync. I added comments too as you requested.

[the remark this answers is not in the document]
Bonifase: Cool, get rid of **noqa in your code, they are not necessary. Use delimiter (“””) for comments that span multiple lines.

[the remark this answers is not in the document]
bevkololi: All done.

api/response/models.py
```diff
@@ -22,7 +22,7 @@ class Response(Base, Utility):
     question_id = Column(Integer, ForeignKey('questions.id'))
     rate = Column(Integer, nullable=True)
     check = Column(Boolean, nullable=True)
-    text_area = Column(String, nullable=True)
+    additional_comment = Column(String, nullable=True)
```
cwizard2011: Why changing this field to additional comment, in the question model, we have three field input, check, and rate. The text_area field is meant to take all input responses, while rate is for rating responses and check for missing_items, Instead of changing the text_area field to additional comment, you can always assign `additional comment` to the `text_area field` in your query implementation.  Other query are using that field which might make the app crash the moment you changed the name. Or what did you think?

api/question/schema_query.py
```diff
@@ -29,12 +34,32 @@ def get_room_response(self, unique_responses, question_id, question_type):
...
+            all_missing_resources = ResourceModel.query.filter(
+                ResourceModel.resources.any(room_id=room.room_id)).count()
+            total_missing_resources = '{} out of {}'.format(
+                all_missing_resources, all_resources)
+            comment = ResponseModel.query.filter_by(
+                room_id=room.room_id).filter(column("additional_comment").isnot(None))  # noqa: E501
+            if comment.count() > 0:
+                last_additional_comment = comment[-1].additional_comment
+            else:
+                last_additional_comment = "No comment yet"
```
Bonifase: What if as a user I give my response in the comment box as "No comment yet", how will we differentiate my comment from this response. Perhaps you could give a more descriptive response to differentiate my comment as a user from this API response.

bevkololi: Noted @Bonifase, I'll get to that immediately.

api/response/models.py
bevkololi: @cwizard2011, the reason was because I feel like it is not descriptive. For example, having the rate field as "integer_field". One does not know at first glance what that field does. It is a very quick fix and the app at the moment is not crashing on my side.

cwizard2011: alright, that is cool

api/response/schema_query.py
```diff
@@ -70,3 +77,20 @@ def resolve_room_response(self, info, room_id):
...
+
+    @Auth.user_roles('Admin')
+    def resolve_all_room_responses(self, info):
+        response = []
+        rooms = RoomModel.query.all()
+        for room in rooms:
+            room_name = room.name
+            total_response = ResponseModel.query.filter_by(
+                room_id=room.id).count()
+            room_response = Response.get_query(info).filter_by(room_id=room.id)
```
RonKbS: I notice you use Response rather than ResponseModel here.
Are they different?

joshuaocero:
LGTM

PR #182, opened 2019-01-07
#162781153 Filter floors by block

caveinn:
**What does this PR do?**
-  Enables an admin to filter floors by block

**Description of task to be completed**
- Ensure that as an admin I can filter the floors by blocks

- Run the `filterByblock` query  passing a `blockId` that is oresent in the datbase as shown below

 ```
 query {
  filterByBlock(blockId: 1){
    id
    name
    blockId
}
}
```

**Screenshots**
image

[#162241777](https://www.pivotaltracker.com/story/show/162781153)

### DIFF
--- a/api/floor/schema.py
+++ b/api/floor/schema.py
@@ -99,6 +99,7 @@ class Query(graphene.ObjectType):
         lambda: Room,
         floor_id=graphene.Int()
     )
+    filter_by_block = graphene.List(Floor, blockId=graphene.Int())
 
     def resolve_all_floors(self, info):
         query = Floor.get_query(info)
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by RonKbS** [comment]
Nice work @caveinn.

Is the Alembic file necessary if no changes to the models that create tables in the database have been introduced?

**Comment by caveinn** [comment]
> Nice work @caveinn.
> 
> Is the Alembic file necessary if no changes to the models that create tables in the database have been introduced?

My tests were failing locally without including them but I have consulted and been informed  that I should ignore it when pushing to the remote.

File: api/floor/schema.py [hunk 1]
```diff
@@ -109,6 +110,14 @@ def resolve_get_rooms_in_a_floor(self, info, floor_id):
         rooms = query.filter(RoomModel.floor_id == floor_id)
         return rooms
 
+    @Auth.user_roles('Admin')
+    def resolve_filter_by_block(self, info, blockId):
+        query = Floor.get_query(info)
+        floors = query.filter_by(block_id=blockId)
+        if not floors:
```
Bonifase: Also, write a test for this case

caveinn: @Bonifase I have implemented the requested change.

Bonifase: Use Andela engineering-playbook as a guide for your PR description

caveinn: Sure, I will get on to that immidietly.

PR #185, opened 2019-01-07
#163009339 Fix room deletion error 

migot01:
#### What does this PR do?
- Adds a responses table

Adds a missing responses table onto the database to enable deletion of rooms. This is currently not possible as a relationship exists between the rooms table and responses table and without the responses table then deleting rooms is not possible

- Run migrations to update the database with the new models
- Run mutations
N/A
https://www.pivotaltracker.com/story/show/163009339
#### Screenshots 
image

#### Questions:


joshuaocero:
I am curious as to why this response table was not there in the first place. In your PR description, you need to provide the background context.

migot01:
@joshuaocero  and @Bonifase  I have implemented the changes requested

PR #188, opened 2019-01-10
#163111433 Enable admin to filter room responses by number of responses

caveinn:
#### What does this PR do?
- Have implementation of filtration of rooms by the number of responses
- Have the implementation of a query that accepts three arguments `filterBy` `upper_limit` and `lower_limit` and returns the rooms whose total response is within the range of the two. The rationale fot the filterBy filed is to allow the same field to be used by another developer in the future who requires to filter the responses by a different metric. I had to change the response schema_query to include the filter since this is where the code for the room responses is. I also introduced a fixture file that contains the data that I will use for testing the query I created. Finally introduced a test file that tests the functionality of the introduced query. This ensures that if any other developer were to break the code it would easily be caught by just running the test.

- After cloning this repo cd into mrm_api
- Checkout to the `ft-filter-by-responses-163111433`  branch 
- Start the server as explained in the `README`.
 - Run the following query 
```
query{
allRoomResponses(filterBy:"Responses",upperLimit: 2, lowerLimit: 0 ){
    responses{
      totalResponses
      roomName
      response{
        responseId
        missingItems
        suggestion
      }
      
    }
}
}
```
- The responses mentioned above are the responses to the questions about the room.
[#163111433](https://www.pivotaltracker.com/n/projects/2154921/stories/163111433)
image

### DIFF
--- a/api/response/schema_query.py
+++ b/api/response/schema_query.py
@@ -29,7 +29,12 @@ class AllResponses(graphene.ObjectType):
 class Query(graphene.ObjectType):
     room_response = graphene.Field(
         RoomResponse, room_id=graphene.Int())
-    all_room_responses = graphene.Field(AllResponses)
+    all_room_responses = graphene.Field(
+        AllResponses,
+        filter_by=graphene.String(),
+        upper_limit=graphene.Int(),
+        lower_limit=graphene.Int()
+    )
 
     def get_room_response(self, test_response):
         response = []
@@ -78,8 +83,7 @@ class Query(graphene.ObjectType):
                 total_responses=total_response,
                 response=responses)
 
-    @Auth.user_roles('Admin')
-    def resolve_all_room_responses(self, info):
+    def get_all_reponses(self, info):
         response = []
         rooms = RoomModel.query.all()
         for room in rooms:
@@ -93,4 +97,29 @@ class Query(graphene.ObjectType):
                 total_responses=total_response,
                 response=all_responses)
             response.append(responses)
-        return AllResponses(responses=response, room_name=room_name)
+        return (response, room_name)
+
+    def filter_rooms_by_responses(self, info, upper_limit, lower_limit):
+        all_responses, room_name = Query.get_all_reponses(self, info)
+        filtered_responses = []
+        for response in all_responses:
+            reponse_count = response.total_responses
+            if lower_limit <= reponse_count <= upper_limit:
+                filtered_responses.append(response)
+        return (filtered_responses, room_name)
+
+    @Auth.user_roles('Admin')
+    def resolve_all_room_responses(
+            self, info, filter_by=None,
+            lower_limit=None, upper_limit=None):
+        responses, room_name = Query.get_all_reponses(self, info)
+        if filter_by == 'Responses':
+            if isinstance(lower_limit, int) and isinstance(upper_limit, int):
+                responses, room_name = Query.filter_rooms_by_responses(
+                    self, info, upper_limit, lower_limit
+                )
+            else:
+                raise GraphQLError("lower_limit and upper_limit are \
+                required to filter by responses")
+
+        return AllResponses(responses=responses, room_name=room_name)
--- a/fixtures/response/room_response_fixture.py
+++ b/fixtures/response/room_response_fixture.py
@@ -88,3 +88,42 @@ summary_room_response_data = {
     }
   }
 }
+
+filter_by_response_query = '''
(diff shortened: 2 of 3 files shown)


api/room/schema_query.py
```diff
@@ -210,3 +217,13 @@ def resolve_bookings_analytics_count(self, info, start_date, end_date):  # noqa:
         analytics = RoomAnalyticsRatios.get_bookings_analytics_count(
             self, query, start_date, end_date)
         return analytics
+
+    def resolve_filter_rooms_by_responses(self, info, lower_bound, upper_bound):
+        query = Room.get_query(info)
+        rooms = query.all()
+        filtered_rooms = []
+        for room in rooms:
```
RonKbS: Great work @C3real-kill3r.

What do you think about placing this `count = ResponseModel.query.filter_by(room_id=room.id).count()` expression outside of the loop.
I believe you only need to retrieve the count once and then use it in the loop.

caveinn: Hello @RonKbS  Actually I need it in every loop. The loop gives me a room and then the count gives me the number of responses for that room. Since I am counting for different rooms that is the reason it is within the loop. Hope that makes sense.

fixtures/room/filter_by_responses_fixture.py
```diff
@@ -0,0 +1,21 @@
+filter_by_responses_query = '''
+query{
+  filterRoomsByResponses(lowerBound:0,upperBound:2){
+    id
+    name
+    roomType
+}
```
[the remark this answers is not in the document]
caveinn: @kampkelly  I have implemented that.

api/room/schema_query.py
Bonifase: Kindly use more descriptive variable names. What does count represent?

fixtures/room/filter_by_responses_fixture.py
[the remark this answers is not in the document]
kampkelly: Okay, great.

caveinn:
@Bonifase  I have implemented the changes you requested

api/question/schema_query.py
```diff
@@ -25,6 +25,12 @@ class Query(graphene.ObjectType):
                                        page=graphene.Int(),
                                        per_page=graphene.Int())
 
+    filter_rooms_by_responses = graphene.Field(
+        Responses,
+        lower_bound=graphene.Int(),
```
cwizard2011: Can you please use a more descriptive variable name? I have been trying hard to understand what you mean by upper bound and lower bound

api/question/schema_query.py
```diff
@@ -25,6 +25,12 @@ class Query(graphene.ObjectType):
...
+        upper_bound=graphene.Int(),
```
cwizard2011: Can you please use a more descriptive variable name? I have been trying hard to understand what you mean by upper bound and lower bound

kampkelly:
This looks good. Just implement @cwizard2011 feedback.

api/question/schema_query.py
```diff
@@ -39,6 +45,33 @@ def get_room_response(self, unique_responses, question_id, question_type):
             response.append(room_response)
         return response
 
+    def resolve_filter_rooms_by_responses(self, info, lower_limit, upper_limit):
+        query = Room.query
+        rooms = query.all()
+        response_query = ResponseModel.query
+        filtered_rooms_responses = []
+        for room in rooms:
+            room_responses = response_query.filter_by(room_id=room.id)
```
Bonifase: What if you just add .count at the end of this line and have only one line for total responses here?

caveinn: @Bonifase I did not include the count here because I also needed the responses themselves to calculate the average rating. If I were to add count there I would still have to add another line for getting all the responses.

api/question/schema_query.py
```diff
@@ -39,6 +45,33 @@ def get_room_response(self, unique_responses, question_id, question_type):
...
+            response_count = room_responses.count()
+            if response_count > 0:
+                average_rating = sum([
```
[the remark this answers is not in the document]
caveinn: @Bonifase  I have implented the feedback.

joshuaocero:
> My opinion is you should filter allRoomResponses query. This will be much easier and most of the functionality is already done here.

You may not even need to touch the backend code. See if the front end guys can use the allRoomResponses endpoint with a graphql filter to get the data they want.

PR #189, opened 2019-01-11
#163066222 make validations to image and name

michael-basweti:
#### What does this PR do?
* Add Validations to imageUrl and location name.
* Added validations that ensure the string put as the imageUrl is a valid url and also the location name should be a valid name like Nairobi or Lagos.
* git checkout to `bg-validate-image-location-name-163066222`, `pip install validators`  and run the following mutation:
```
 mutation {
  createLocation(name: "Kampala", abbreviation: "KLA", country: "Uganda", timeZone: "EAST_AFRICA_TIME") {   # noqa E501
    location {
      name
    }
  }
}
```
```
mutation{
    updateLocation(locationId: 1, name: "Nairobi", country: "Kenya", abbreviation: "KE",imageUrl:"http://image.com"){ # noqa: E501
        location{
            name
            abbreviation
            country
        }
    }
}
```
* play with name and imageUrl in updateLocation mutation to check what is valid
* Not applicable
#### screenshots:
* invalid url
screenshot 2019-01-11 at 09 50 28
* invalid location name
screenshot 2019-01-11 at 09 51 00
* everything correct
screenshot 2019-01-11 at 09 51 21

[#163066222](https://www.pivotaltracker.com/n/projects/2154921/stories/163066222)

### DIFF
--- a/utilities/utility.py
+++ b/utilities/utility.py
@@ -1,5 +1,16 @@
 import datetime
+import validators
 from helpers.database import db_session
+
+
+def validate_url(**kwargs):
+    """
+    Function to validate url when
+    saving an object
+    :params kwargs
+    """
+    if not validators.url(kwargs.get('image_url')):
+        raise AttributeError("Please enter a valid image url")
 
 
 def validate_empty_fields(**kwargs):
--- a/api/location/schema.py
+++ b/api/location/schema.py
@@ -6,7 +6,7 @@ from api.location.models import Location as LocationModel
 from api.devices.models import Devices as DevicesModel  # noqa: F401
 from api.room_resource.models import Resource as ResourceModel  # noqa: F401
 from api.room.schema import Room
-from utilities.utility import validate_country_field, validate_timezone_field, update_entity_fields, validate_empty_fields  # noqa: E501
+from utilities.utility import validate_country_field, validate_timezone_field, update_entity_fields, validate_empty_fields, validate_url  # noqa: E501
 from helpers.room_filter.room_filter import room_join_location
 
 
@@ -28,6 +28,7 @@ class CreateLocation(graphene.Mutation):
         # Validate if the country given is a valid country
         validate_country_field(**kwargs)
         validate_timezone_field(**kwargs)
+        validate_url(**kwargs)
         location = LocationModel(**kwargs)
         location.save()
         return CreateLocation(location=location)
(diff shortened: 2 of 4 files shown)


utilities/utility.py
```diff
@@ -29,6 +40,21 @@ def validate_country_field(**kwargs):
...
 
 
+def validate_name_field(**kwargs):
+    """
+    Function to validate location fields when
+    saving an object
+    :params kwargs
+    """
+    my_args = []
+    locations = ['Lagos', 'Kampala', 'Nairobi']
```
C3real-kill3r: Good work @michael-basweti 
Jus a clarification I need, I was wondering since Rwanda opened recently should it also be included on the list of locations, and if there is any rationale for omitting it kindly lemme know. Cheers!

utilities/utility.py
```diff
@@ -1,7 +1,18 @@
 import datetime
+import validators
 from helpers.database import db_session
 
 
+def validate_url(**kwargs):
+    """
+    Function to validate empty fields when
+    saving an object
+    :params kwargs
```
kampkelly: Nice work @michael-basweti.
But the documentation for this function is not right yet. It is referring to empty fields instead of image url. 
Also consider changing the function name to `validate_image_url_field`. This may be more clearer and descriptive.

utilities/utility.py
```diff
@@ -29,6 +40,21 @@ def validate_country_field(**kwargs):
         raise AttributeError("Not a valid country")
 
 
+def validate_name_field(**kwargs):
+    """
+    Function to validate location fields when
+    saving an object
+    :params kwargs
```
kampkelly: @michael-basweti  You can also consider using `validate_location_name_field` here.

utilities/utility.py
michael-basweti: Will change that right away, thanks

utilities/utility.py
michael-basweti: Rwanda has not been added to Converge yet, will be added later

utilities/utility.py
michael-basweti: I just realised that you are referring to the wrong function, that was there before

utilities/utility.py
```diff
@@ -1,7 +1,18 @@
...
 
 
+def validate_url(**kwargs):
+    """
+    Function to validate url when
+    saving an object
+    :params kwargs
+    """
+    if not validators.url(kwargs.get('image_url')):
+        raise AttributeError("the url should be valid")
```
Bonifase: I would suggest using a more specific message like "The image url should be valid" or something like "Please enter a valid image url" same to the location name.

utilities/utility.py
michael-basweti: @kampkelly , the validation can be used to validate other urls so I think it should not be named after a specific field.

utilities/utility.py
joshuaocero: @michael-basweti the logic here is not clear to me. What does the `my_args` variable contain?

utilities/utility.py
kampkelly: Okay. Got it.

PR #191, opened 2019-01-14
#163143276 Paginate the wings response

kampkelly:
#### What does this PR do?
- Paginate the wings response

The task is to allow the admin to be able to scroll through all the available wings .

- Run the following query
```
query {
    allWings(page: 1, perPage: 2) {
        wings {
            id
            name
            floorId
        }
        queryTotal
        pages
        hasNext
        hasPrevious
    }
}
```

None

[#163143276](https://www.pivotaltracker.com/story/show/163143276)

#### Screenshots
screen shot 2019-01-14 at 11 53 08 am

**Questions:**
None

### DIFF
--- a/api/wing/schema.py
+++ b/api/wing/schema.py
@@ -8,6 +8,7 @@ from utilities.utility import update_entity_fields, validate_empty_fields
 from helpers.auth.authentication import Auth
 from helpers.auth.error_handler import SaveContextManager
 from helpers.auth.admin_roles import admin_roles
+from helpers.pagination.paginate import Paginate, validate_page
 
 
 class Wing(SQLAlchemyObjectType):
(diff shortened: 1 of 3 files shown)


api/wing/schema.py
```diff
@@ -73,12 +74,37 @@ def mutate(self, info, wing_id, **kwargs):
...
         query = Wing.get_query(info)
-        return query.all()
+        if not page:
+            return query.all()
+        page = validate_page(page)
+        self.query_total = query.count()
+        result = query.limit(per_page).offset(page*per_page)
+        if result.count() == 0:
+            return GraphQLError("No more resources")
+        return result
```
caveinn: Isn't it better to raise an error rather than return an error?

bevkololi: +1

api/wing/schema.py
```diff
@@ -73,12 +74,37 @@ def mutate(self, info, wing_id, **kwargs):
...
+        per_page = self.per_page
         query = Wing.get_query(info)
-        return query.all()
+        if not page:
+            return query.all()
+        page = validate_page(page)
+        self.query_total = query.count()
+        result = query.limit(per_page).offset(page*per_page)
+        if result.count() == 0:
+            return GraphQLError("No more resources")
```
cwizard2011: Please implement the feedback given by @bevkololi and @caveinn , `raise GraphQLError`

cwizard2011: Can you use a more descriptive message here? something like `No wing on this page` or something better, considering we also have resources schema in the application, this error message can be confusing

api/wing/schema.py
kampkelly: This has been changed now. Thanks.

api/wing/schema.py
```diff
@@ -73,12 +74,37 @@ def mutate(self, info, wing_id, **kwargs):
...
+        per_page = self.per_page
         query = Wing.get_query(info)
-        return query.all()
+        if not page:
+            return query.all()
+        page = validate_page(page)
+        self.query_total = query.count()
+        result = query.limit(per_page).offset(page*per_page)
+        if result.count() == 0:
+            raise GraphQLError("No more resources")
```
cwizard2011: Can you use a more descriptive message here? something like No wing on this page or something better, considering we also have resources schema in the application, this error message can be confusing

kampkelly: @cwizard2011 I agree with you. This has now been changed to `No more wings` as it is more descriptive.

api/wing/schema.py
kampkelly: The feedback has been implemented.

joshuaocero:
This feature is no longer necessary as we shall be restructuring the room structure to have wing information as tags

PR #192, opened 2019-01-16
#163187782: Merge feedback and response entities

cwizard2011:
#### What does this PR do?
- Merged the feedback and response entities to remove duplication

- Remove feedback model and all its associations in user and room table
- Remove feedback schema and update schema.py file in the root directory
- Create a new migration to drop feedback table in the database
- Remove all fixtures and tests associated with the feedback table

- The feedback and the response entity achieve the same goal, i.e allowing users to give their feedback in a meeting room, the feedback entity was initially created for allowing users give random feedback about the meeting room, after some months the feedback feature was replaced by the feedback initiatives where the users will be able to respond to feedback questions, hence the need to remove the old feedback entity since the new response entity already achieve the goal of the feedback entity

[#163187782](https://www.pivotaltracker.com/story/show/163187782)


kampkelly:
Looks good to me.

PR #193, opened 2019-01-18
#163283090: Refactor validations on country field and timezone field

bevkololi:
#### What does this PR do?
Refactor `validate_country_field` and `validate_timezone_field`  function in the Utility file to the Location Model.

- Remove validations on country field and timezone field from utility file. This is because of import errors I faced in the utility file. 
- Created new file in utility folder and refactored the above validations to be used in creating and updating locations. This included removing the hardcoded values.

- The validations basically check if the values are of type Enum

#### Screenshots
screenshot 2019-01-18 at 12 14 50

[#163283090](https://www.pivotaltracker.com/story/show/163283090)


bevkololi:
> Kindly share the reason for removing the validate_url function because it seems to be serving a purpose different from that of validating country field and timezone field.

This is an error on my end. Fixing it now. Thank you @RonKbS

utilities/validations.py
```diff
@@ -0,0 +1,27 @@
...
+    valid_countries = [country.value for country in CountryType]
+    country_name = kwargs['country']
+    if country_name not in valid_countries:
+        raise AttributeError("Not a valid country")
+
+
+def validate_timezone_field(**kwargs):
+    """
+    Function to validate country fields when
+    saving an object
```
kampkelly: @bevkololi please correct the documentation here.  It should be to validate timezone fields, not country fields. 
Good work though.

bevkololi: @kampkelly, thanks, correcting this now.

api/location/schema.py
```diff
@@ -68,15 +68,8 @@ def mutate(self, info, location_id, **kwargs):
             validate_timezone_field(**kwargs)
         if "country" in kwargs:
             validate_country_field(**kwargs)
-        if "image_url" in kwargs:
-            validate_url(**kwargs)
-        if "abbreviation" in kwargs:  # noqa
+        if "name" in kwargs or "abbreviation" in kwargs or "imageUrl" in kwargs:  # noqa
```
joshuaocero: @bevkololi what does this if statement do? What is it checking for?

joshuaocero:
I like that you created a new file for validations. How did you select which functions go to the new utility file and which ones stay? Please shed some light on the logic behind that separation.

bevkololi:
> @bevkololi I like that you created a new file for validations. How did you select which functions go to the new utility file and which ones stay? Please shed some light on the logic behind that separation.

@joshuaocero , since the task involved validation from the model, it required imports from other files. However we stumbled upon a blocker whereby imports from models in the utility file brought an import error. After a sync with the backend team, I learnt that this was inbuilt in Flask i.e one couldn't import from models in the utility file in flask. The new file therefore accommodates validations from the models while the rest do not.

api/location/schema.py
bevkololi: I did not have the latest changes from develop. The above lines were removed. The PR is now updated.

utilities/validations.py
```diff
@@ -0,0 +1,27 @@
+from api.location.models import CountryType, TimeZoneType
```
Bonifase: Since we have various validations in the codebase other than for a country field and timezone, the file name would be misleading if it only contains these two methods unless you are planning to move all the existing validation methods into this new file.

bevkololi: Yes I concur. Let me move all of them into this file.

Bonifase: Or you could just rename it to what it specifically does

bevkololi: Yes I get you. The initial thought was create this file for validations of data drawn from models. I'll move the validations into this file to prevent the confusion.

kampkelly:
LGTM

PR #194, opened 2019-01-18
#160609522 Record when room-feedback-questions are viewed

RonKbS:
#### What does this PR do?
The PR enables the converge team keep track of the number of views on the feedback questions

Currently, the team does not know if people using the rooms view the questions but do not send back any feedback.  The PR raised seeks to remedy that by introducing a counter that increments everytime the `feedback` button is pressed on the Android platform.

- Run the following query.
```    mutation{
  updateQuestionViews(incrementTotalViews:true) {
    questions{
      id
      question
      totalViews
      }
  }
}
```

#### Any background context you want to add?
- The PR answers the request of the Fellow Project Coordinator to see those who view the feedback questions but do not give any feedback.

#### Screenshots
screenshot 2019-01-25 at 12 40 47

[#162576320](https://www.pivotaltracker.com/story/show/162576320)

### DIFF
--- a/fixtures/questions/create_questions_fixtures.py
+++ b/fixtures/questions/create_questions_fixtures.py
@@ -164,3 +164,15 @@ all_questions_response = {
         ]
     }
 }
+
+query_update_total_views_of_questions = '''
+mutation{
+  updateQuestionViews(incrementTotalViews:true) {
+    questions{
+      question
+      questionType
+      totalViews
+      }
+  }
+}
+'''
--- a/api/question/schema.py
+++ b/api/question/schema.py
@@ -118,7 +118,28 @@ class DeleteQuestion(graphene.Mutation):
         return DeleteQuestion(question=exact_question)
 
 
+class UpdateQuestionViews(graphene.Mutation):
+    class Arguments:
+        increment_total_views = graphene.Boolean(required=True)
+
+    questions = graphene.List(Question)
+
+    def mutate(self, info, **kwargs):
+        query = Question.get_query(info)
+        questions = query.all()
+        new_total_views = 0
+        for question in questions:
+            if kwargs['increment_total_views'] and not question.total_views:
+                new_total_views = 1
+            if kwargs['increment_total_views'] and question.total_views:
+                new_total_views = question.total_views + 1
+            update_entity_fields(question, total_views=new_total_views)
+            question.save()
+        return UpdateQuestionViews(questions=questions)
+
+
 class Mutation(graphene.ObjectType):
     create_question = CreateQuestion.Field()
     delete_question = DeleteQuestion.Field()
     update_question = UpdateQuestion.Field()
+    update_question_views = UpdateQuestionViews.Field()
--- a/tests/test_questions/test_update_question.py
+++ b/tests/test_questions/test_update_question.py
@@ -4,7 +4,8 @@ from tests.base import BaseTestCase, CommonTestCases
 from fixtures.questions.create_questions_fixtures import (
    update_question_mutation,
    update_question_response,
-   update_question_invalidId
+   update_question_invalidId,
+   query_update_total_views_of_questions
 )
 
 
@@ -32,4 +33,21 @@ class TestUpdateQuestion(BaseTestCase):
             self,
             update_question_invalidId,
             "Question not found"
+        )
+
+    def test_increment_total_views_of_questions(self):
+        """
+        Testing for incrementing the number of views
+        for a room's feedback qustion
+
+        """
+        CommonTestCases.user_token_assert_in(
+            self,
+            query_update_total_views_of_questions,
+            '1'
+        )
+        CommonTestCases.user_token_assert_in(
+            self,
+            query_update_total_views_of_questions,
+            '2'
         )


api/room/schema.py
```diff
@@ -147,7 +147,32 @@ def mutate(self, info, room_id, **kwargs):
...
+        validate_empty_fields(**kwargs)
+        query = Room.get_query(info)
+        exact_room = query.filter_by(id=kwargs['room_id']).first()
+        if not exact_room:
+            raise GraphQLError("Non-existent room id")
+
+        if exact_room.total_room_feedback_views:
+            new_total_feedback_view = exact_room.total_room_feedback_views + 1
+        if not exact_room.total_room_feedback_views:
+            new_total_feedback_view = 1
```
bevkololi: Is there a reason for this? Why not just see 0 as the total number of views?

api/room/schema.py
```diff
@@ -147,7 +147,32 @@ def mutate(self, info, room_id, **kwargs):
...
+    def mutate(self, info, **kwargs):
+        validate_empty_fields(**kwargs)
+        query = Room.get_query(info)
+        exact_room = query.filter_by(id=kwargs['room_id']).first()
+        if not exact_room:
+            raise GraphQLError("Non-existent room id")
+
+        if exact_room.total_room_feedback_views:
+            new_total_feedback_view = exact_room.total_room_feedback_views + 1
+        if not exact_room.total_room_feedback_views:
```
bevkololi: Also, why use if not instead of if...else?

api/room/schema.py
RonKbS: Initially, the DB has NULL set for this field and I looked up ways of changing this to a value of 0 such that I could do the incrementing from the start. I was not able to find one unfortunately hence the second if-statement.

Context: Addition cannot be done on `exact_room.total_room_feedback_views` when it is NULL.

api/room/schema.py
```diff
@@ -147,7 +147,32 @@ def mutate(self, info, room_id, **kwargs):
...
+
+    class Arguments:
+        room_id = graphene.Int(required=True)
+    room = graphene.Field(Room)
+
+    def mutate(self, info, **kwargs):
+        validate_empty_fields(**kwargs)
+        query = Room.get_query(info)
+        exact_room = query.filter_by(id=kwargs['room_id']).first()
+        if not exact_room:
```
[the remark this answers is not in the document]
RonKbS: Yes, that would reduce the repetition.
It might need to be raised in another PR though so that the scope is not exceeded for this one.

api/room/schema.py
kampkelly: @bevkololi I think what this code is doing is to increment the number of views. So that is why he is adding a 1 to the number of views in the first if statement and setting it to 1 in the second if statement since the current number of views there is 0 or NULL.

bevkololi: Got it @kampkelly, @RonKbS

api/room/schema.py
```diff
@@ -147,7 +147,32 @@ def mutate(self, info, room_id, **kwargs):
...
+        )
+        exact_room.save()
+        return IncrementTotalRoomFeedbackView(room=exact_room)
+
+
 class Mutation(graphene.ObjectType):
     create_room = CreateRoom.Field()
     update_room = UpdateRoom.Field()
     delete_room = DeleteRoom.Field()
+    increment_room_feedback = IncrementTotalRoomFeedbackView.Field()
```
joshuaocero: @RonKbS any reason why you did not just use the update_room mutation?

RonKbS: I have looked through the two pieces of code and realized there is indeed a similarity that would point to me having to include these changes in the update_room mutation.

However, the idea of incrementing responses (for a room's feedback question) is separate from updating a room I think, so I hesitate to do this. I will defer to your decision on the matter.

omobosteven: @RonKBS except for obvious reasons, I think this should go to the update_room_mutation. You're actually editing a resource in the room, it doestn't matter if it is a different story/task.

Bonifase: Definitely, this should be part of updating room details, I also think that the admin should be able to see the total number of feedback views when they query a particular room details.

RonKbS: Alright,

I have begun working on an implementation that puts these changes in the `update_room` mutation.

RonKbS:
>Definitely, this should be part of updating room details, I also think that the admin should be able to see the total number of feedback views when they query particular room details.

@Bonifase I have implemented the recommended changes. The number of feedback views can be queried for by adding `totalRoomFeedbackViews` to queries for rooms.

api/room/schema.py
```diff
@@ -124,6 +125,19 @@ def mutate(self, info, room_id, **kwargs):
         admin_roles.update_delete_rooms_create_resource(room_id)
         update_entity_fields(exact_room, **kwargs)
 
+        new_total_feedback_views = 0
+        if 'increment_room_feedback_views' in kwargs and\
```
Bonifase: Could you kindly explain what this line of code is doing?

RonKbS: The code checks if the argument for `increment_room_feedback_views` is in the `updateRoom`  mutation. Using the mutation for updating other fields of the room will serve an error if the argument and that line are excluded.

Bonifase: I mean `and\` part of the line

Bonifase: I mean and\ part of the line

RonKbS: Rather than have `noqa`, I use the backslash `\` to split the line of code into three. The `and` is to ensure all the conditions are met before the block within the `if-statement` is executed.

Bonifase: Let's have a better way of doing this, the implementation would make the code difficult to read for future developers.

Bonifase: I would suggest that you configure your editor to autoformat by limiting 79 characters per line.

RonKbS: This current implementation is to make sure the lines do not exceed the 79 limit. To break a line of code that is not a string, the `\` is required. Would it be okay to add a comment explaining the multiple conditions in one if statement to help future developers?

alembic/versions/5e09fc587903_add_total_room_feedback_views_column.py
```diff
@@ -0,0 +1,28 @@
+"""add total_room_feedback_views_column
```
Bonifase: Ensure to run a fresh migration from the develop branch to avoid multiple head errors

RonKbS: I have updated the alembic file @Bonifase.

RonKbS:
> Fix failing tests

The tests have been fixed @Bonifase

dnuwa:
Looks good

fixtures/room/create_room_fixtures.py
```diff
@@ -280,3 +280,31 @@
             "createRoom": null
         }
     }
+
+increment_total_room_feedback_mutation = '''
```
Bonifase: Is it increment total room feedback or total room feedback views?

RonKbS: These changes are part of the old implementation so let me remove them since they are not part of the new implementation

fixtures/questions/create_questions_fixtures.py
```diff
@@ -164,3 +164,15 @@
         ]
     }
 }
+
+query_update_total_views = '''
```
Bonifase: Kindly use a more descrip[tive variable names. With this name one cannot tell 'query_update_total_views' is the total views for what exactly.

RonKbS:
> Fix the failing tests

The tests have passed @Bonifase

dnuwa:
LGTM

PR #201, opened 2019-01-23
#163294072 Add room tags model

kampkelly:
#### What does this PR do?
- Creates the room tags model

The task is to create a room tags model. This is to allow for a uniform office structure across all locations.

- Clone the repo.
- Run `make build`
- `make run-app`
- Then run migrations
- You will see tags and room_tags tables added to the database.

You may need to connect Postico to your database to view the created tables.

[#163294072](https://www.pivotaltracker.com/story/show/163294072)

#### Screenshots

**Questions:**
None


alembic/env.py
```diff
@@ -39,6 +39,7 @@
 from api.events.models import Events
 from api.question.models import Question
 from api.response.models import Response
+from api.tag.models import Tag
```
RonKbS: Great work @kampkelly  👍 .

Can you share with us why this line was included?

kampkelly: I am importing the tag model here so alembic can generate migrations for the tags whenever there is a change in the model.

api/room/models.py
```diff
@@ -21,6 +21,7 @@ class Room(Base, Utility):
     image_url = Column(String)
     calendar_id = Column(String)
     floor_id = Column(Integer, ForeignKey('floors.id'))
+    tag_id = Column(Integer, ForeignKey('tags.id'))
```
Bonifase: Is it possible to have one room having more than one tag?

api/tag/models.py
```diff
@@ -0,0 +1,33 @@
+from sqlalchemy import (Column, String, Table, Integer, ForeignKey)
+from sqlalchemy.orm import relationship
+# from sqlalchemy.schema import Sequence
+
+from helpers.database import Base
+from utilities.utility import Utility, validate_empty_fields
+
+room_tags = Table(
```
Bonifase: Is there a reason why you have this variable here to define the relationship instead of defining the relationship directly in the Tag class?

tests/base.py
```diff
@@ -79,6 +80,10 @@ def setUp(self):
             wing.save()
             wing_two = Wing(name="Big Apple", floor_id=floor_two.id)
             wing_two.save()
+            tag = Tag(name='Block-A',
+                        color='green',
+                        description='The description')  # noqa: E501
```
Bonifase: Please avoid the use of noqa in the codebase as much as possible.

api/room/models.py
kampkelly: Yes, a room can have more than one tag. This is done in the room_tags table where several room id's can have the same tag id.
Looking at this again, I see there is no need to have a tag_id in the room models. This will be removed shortly.

tests/base.py
kampkelly: Thanks for the feedback. Noqa has been removed and the line modified.

api/room/models.py
kampkelly: The tag id has now been removed from the room model.

api/tag/models.py
kampkelly: Yes, I followed the implementation as described in the sqlalchemy documentation here: [https://docs.sqlalchemy.org/en/latest/orm/basic_relationships.html#many-to-many](https://docs.sqlalchemy.org/en/latest/orm/basic_relationships.html#many-to-many).
However, I now made a small change to use the variable name (room_tags) in the secondary argument of the relationship instead of room_tags as a string.
It should be noted that both the variable name and the string name seems to work just fine.

PR #204, opened 2019-01-24
#163189549 Implement error handling for missing resource

C3real-kill3r:
#### What does this PR do?
* Handle all 404 error message when the resource for bookings analytics counts is not found.
*  Added an exception handler to the `analytics_helper.py` file
* git checkout to `ch-handle-HTTP-errors-163189549` and run the following query:
* Run the `bookingsAnalyticsCount` query with `startDate` and `endDate`(both are required fields).

* Not applicable
#### screenshots:

* previous error message when a wrong calendar id is supplied

screenshot 2019-01-24 at 17 21 46

* new error message when a wrong calendar id is supplied

screenshot 2019-01-24 at 17 22 46

[#163189549](https://www.pivotaltracker.com/story/show/163189549)

### DIFF
--- a/helpers/calendar/analytics_helper.py
+++ b/helpers/calendar/analytics_helper.py
@@ -85,9 +85,12 @@ class CommonAnalytics(Credentials):
             - min_limit, max_limit(Time range)
         """
         service = Credentials.set_api_credentials(self)
-        events_result = service.events().list(
-            calendarId=calendar_id, timeMin=min_limit, timeMax=max_limit,
-            singleEvents=True, orderBy='startTime').execute()
+        try:
+            events_result = service.events().list(
+                calendarId=calendar_id, timeMin=min_limit, timeMax=max_limit,
+                singleEvents=True, orderBy='startTime').execute()
+        except Exception:
+            raise GraphQLError("Resource not found")
         calendar_events = events_result.get('items', [])
         return calendar_events
 


helpers/calendar/analytics_helper.py
```diff
@@ -85,9 +85,12 @@ def get_all_events_in_a_room(self, calendar_id, min_limit, max_limit):
...
         service = Credentials.set_api_credentials(self)
-        events_result = service.events().list(
-            calendarId=calendar_id, timeMin=min_limit, timeMax=max_limit,
-            singleEvents=True, orderBy='startTime').execute()
+        try:
+            events_result = service.events().list(
+                calendarId=calendar_id, timeMin=min_limit, timeMax=max_limit,
+                singleEvents=True, orderBy='startTime').execute()
+        except Exception:
+            raise GraphQLError("Resource not found")
```
caveinn: Can this be done by catching the specific error instead of a broad error.

joshuaocero: @C3real-kill3r please respond to caveinn's comment. Is there a specific error you are trying to catch?

C3real-kill3r: Sorry about this. The error I was trying to catch was an HTTPError. But when I declared it explicitly, it was failing. I also tried to catch `urllib2.HTTPError` handle it, and if it's Error 404, raise that exception; this too failed to work. So the only option I had left after extensive research was to use a bare except (is a flake8 error) or use the exception class since all built-in, non-system-exiting exceptions are derived from this class; and this got it to work, unless you have any suggestions?

caveinn: It seems you have already tried out all the suggestions I had.

PR #206, opened 2019-01-25
#163219021 Add current page on paginator

nakatuddesuzan:
#### What does this PR do?
 Add current page on paginator
### Description of Task to be completed?
- The current page number is added to the paginator
- Modify the fixtures to cover the added lines
- checkout branch `ft-add-current-page-163219021`
- Run  the application using `make run-app`
- Query any paginated data say rooms.
The current paginator indicates whether a page has_next or has_previous but does not indicate the current page. 
[#163219021](https://www.pivotaltracker.com/story/show/163219021)
###### Valid page query

screen shot 2019-01-28 at 13 46 28

###### Invalid page query

screen shot 2019-01-28 at 13 48 17

### DIFF
--- a/helpers/pagination/paginate.py
+++ b/helpers/pagination/paginate.py
@@ -10,6 +10,7 @@ class Paginate(graphene.ObjectType):
     query_total = graphene.Int()
     has_next = graphene.Boolean()
     has_previous = graphene.Boolean()
+    current_page = graphene.Int()
 
     def __init__(self, **kwargs):
         self.page = kwargs.pop('page', None)
@@ -47,6 +48,12 @@ class Paginate(graphene.ObjectType):
                 has_previous = False
 
         return has_previous
+
+    def resolve_current_page(self, pages):
+        pages = self.resolve_pages(pages)
+        if self.page > pages:
+            raise GraphQLError("Page does not exist")
+        return self.page
 
 
 def validate_page(page):
--- a/fixtures/floor/get_floors_fixures.py
+++ b/fixtures/floor/get_floors_fixures.py
@@ -1,3 +1,4 @@
+null = None
 get_all_floors_query = '''{
     allFloors {
         floors {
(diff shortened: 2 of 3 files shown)


helpers/pagination/paginate.py
```diff
@@ -48,6 +49,12 @@ def resolve_has_previous(self, has_previous):
 
         return has_previous
 
+    def resolve_current_page(self, pages):
+        pages = self.resolve_pages(pages)
+        if self.page > pages:
+            return pages
+        return self.page
```
kampkelly: I see you return the total number of pages as the current page when a user sends a query with a page number that doesn't exist. Won't it be better to return a null or false instead for the current page since it is not actually returning a valid page?
Nice work though @nakatuddesuzan.

caveinn: If I understand correctly this works such that when the page passed is greater than the last page then just send them to the last page.

nakatuddesuzan: Yes @kampkelly , What @caveinn  has pointed is the logic I used.

nakatuddesuzan: The user is sent to the last page. The last page == total number of pages. So the page being returned is valid.

kampkelly: Okay.

Bonifase: If there are no more resources, do we need to return the last page of the resources as the current page? For instance, would this mean that the page 100 == page 7 as in the screenshot above?

Bonifase: If there are no more resources, do we need to return the last page of the resources as the current page? For instance, would this mean that the page 100 == page 7 as in the screenshot above?

nakatuddesuzan: Okay, I now get the argument.  Let me return null for the current page. @Bonifase  and @kampkelly

nakatuddesuzan: This feedback has been worked on @Bonifase

PR #210, opened 2019-01-27
#163456848 Enable admin to view the app bookings percentage

kampkelly:
#### What does this PR do?
- Enables the admin to see app bookings percentage

The task is to admin to view the percentage of bookings that were made from the mobile app. 

- Clone the repo.
- Run `make build`
- `make run-app`
- Then run either of the following queries:
```
query {
  analyticsRatios(startDate: "Dec 09 2018", endDate: "Jan 17 2019") {
     checkins
     cancellations
     bookings
     checkinsPercentage
     roomName
     cancellationsPercentage
     appBookings
     appBookingsPercentage
  }
}
```
#### OR

```
query {
  analyticsRatiosPerRoom(startDate: "Dec 09 2018", endDate: "Jan 17 2019") {
       ratios {
        bookings
         checkins
         cancellations
         cancellationsPercentage
         checkinsPercentage
         appBookingsPercentage
         appBookings
         roomName
      }
  }
}
```
- You will see the app bookings percentage returned.

None

[#163456848](https://www.pivotaltracker.com/story/show/163456848)

#### Screenshots
screen shot 2019-01-28 at 10 30 58 pm
screen shot 2019-01-28 at 10 32 15 pm

**Questions:**
@Bonifase 
(1) Should the number of cancellations be added to the app_bookings?

screen shot 2019-01-29 at 11 42 51 am

### DIFF
--- a/helpers/calendar/analytics_helper.py
+++ b/helpers/calendar/analytics_helper.py
@@ -91,8 +91,8 @@ class CommonAnalytics(Credentials):
                 singleEvents=True, orderBy='startTime').execute()
         except Exception:
             raise GraphQLError("Resource not found")
-        calendar_events = events_result.get('items', [])
-        return calendar_events
+        all_events = events_result.get('items', [])
+        return all_events
 
     def get_event_details(self, event, calendar_id):
         """ Filter details of an event
@@ -111,6 +111,11 @@ class CommonAnalytics(Credentials):
                     event_details["roomName"] = resource.get(
                         'displayName') or None
                     event_details["summary"] = event.get("summary")
+        elif event.get('organizer') and event.get(
+                'organizer').get('email') == calendar_id:
+            event_details["roomName"] = event.get(
+                'organizer').get('displayName') or None
+            event_details["summary"] = event.get("summary")
         return event_details
 
     def get_room_statistics(self, number_of_events_in_room, all_details):
@@ -151,10 +156,10 @@ class CommonAnalytics(Credentials):
         bookings = 0
         rooms = CommonAnalytics.get_calendar_id_name(self, query)
         for room in rooms:
-            calendar_events = CommonAnalytics.get_all_events_in_a_room(
+            all_events = CommonAnalytics.get_all_events_in_a_room(
                 self, room["calendar_id"], start_date, end_date)
-            if calendar_events:
-                bookings += len(calendar_events)
+            if all_events:
+                bookings += len(all_events)
         return bookings
 
     @staticmethod
--- a/helpers/calendar/ratios_and_utilization.py
+++ b/helpers/calendar/ratios_and_utilization.py
@@ -26,22 +26,24 @@ class RoomAnalyticsRatios(Credentials):
             self, query)
 
         checkins = 0
-        events_list = []
         cancellations = 0
+        calendar_events_list = []
+        app_events_list = []
 
         for room in rooms:
-            checkins, cancellations, events_list = RoomAnalyticsRatios.retrieve_cancellations_and_checkins_for_room(self,  # noqa
-                                                        room['calendar_id'],
+            checkins, cancellations, calendar_events_list, app_events_list = RoomAnalyticsRatios().retrieve_cancellations_and_checkins_for_room(room['calendar_id'], # noqa 
                                                         start_date,
                                                         day_after_end_date,
                                                         checkins,
                                                         cancellations,
-                                                        events_list)
+                                                        calendar_events_list,
+                                                        app_events_list)
 
-        ratio_object = RoomAnalyticsRatios.map_results_to_ratio_class(self,  # noqa
+        ratio_object = RoomAnalyticsRatios().map_results_to_ratio_class(
                             checkins,
                             cancellations,
-                            events_list)
+                            calendar_events_list,
+                            app_events_list)
 
         return ratio_object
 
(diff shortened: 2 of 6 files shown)


helpers/calendar/analytics_helper.py
```diff
@@ -108,6 +108,11 @@ def get_event_details(self, event, calendar_id):
                     event_details["roomName"] = resource.get(
                         'displayName') or None
                     event_details["summary"] = event.get("summary")
+        else:
+            if event.get('organizer'):
```
bevkololi: This could work with elif instead of separating them like this

helpers/calendar/ratios_and_utilization.py
```diff
@@ -60,53 +66,68 @@ def get_analytics_ratios_per_room(self, query, start, end):
...
                 checkins,cancellations, output
         """
-        calendar_events = CommonAnalytics.get_all_events_in_a_room(
+        all_events = CommonAnalytics.get_all_events_in_a_room(
                 self, calendar_id, start_date, day_after_end_date)
-        if calendar_events:
-            for event in calendar_events:
-                if event.get('attendees'):
-                    event_details = CommonAnalytics.get_event_details(
+        if all_events:
```
bevkololi: This block has too many if statements i.e it is deeply nested. kindly consider refactoring it to avoid this.

helpers/calendar/analytics_helper.py
```diff
@@ -88,8 +88,8 @@ def get_all_events_in_a_room(self, calendar_id, min_limit, max_limit):
         events_result = service.events().list(
             calendarId=calendar_id, timeMin=min_limit, timeMax=max_limit,
             singleEvents=True, orderBy='startTime').execute()
-        calendar_events = events_result.get('items', [])
-        return calendar_events
+        all_events = events_result.get('items', [])
+        return all_events
```
kampkelly: This has been renamed to all_events because the events returned here is the list of all events not only the events booked from the calendar

helpers/calendar/analytics_helper.py
```diff
@@ -108,6 +108,11 @@ def get_event_details(self, event, calendar_id):
                     event_details["roomName"] = resource.get(
                         'displayName') or None
                     event_details["summary"] = event.get("summary")
+        elif event.get('organizer') and event.get(
+                'organizer').get('email') == calendar_id:
+            event_details["roomName"] = event.get(
+                'organizer').get('displayName') or None
+            event_details["summary"] = event.get("summary")
```
kampkelly: This code was added to also get the event details of those booked from the app. Previously, it was only getting the event details of those booked from the calendar.

helpers/calendar/ratios_and_utilization.py
```diff
@@ -26,22 +26,25 @@ def get_analytics_ratios(self, query, start, end):
...
 
         checkins = 0
-        events_list = []
         cancellations = 0
+        calendar_events_list = []
+        app_events_list = []
 
         for room in rooms:
-            checkins, cancellations, events_list = RoomAnalyticsRatios.retrieve_cancellations_and_checkins_for_room(self,  # noqa
+            checkins, cancellations, calendar_events_list, app_events_list = RoomAnalyticsRatios.retrieve_cancellations_and_checkins_for_room(self,  # noqa
```
caveinn: Try to avoid using noqa. Also instead of passing in self i believe you should create an object of that class before making the method call.

helpers/calendar/ratios_and_utilization.py
kampkelly: This has been refactored.

helpers/calendar/ratios_and_utilization.py
```diff
@@ -57,56 +60,79 @@ def get_analytics_ratios_per_room(self, query, start, end):
...
         response = []
         for room in rooms:
-            events_list = []
             checkins = 0
             cancellations = 0
+            calendar_events_list = []
+            app_events_list = []
 
-            checkins, cancellations, events_list = RoomAnalyticsRatios.retrieve_cancellations_and_checkins_for_room(self,  # noqa
+            checkins, cancellations, calendar_events_list, app_events_list = RoomAnalyticsRatios.retrieve_cancellations_and_checkins_for_room(self,  # noqa
```
caveinn: The self here is also unnecessary if you create the object before calling the method.

helpers/calendar/analytics_helper.py
kampkelly: Done

helpers/calendar/ratios_and_utilization.py
kampkelly: This answer also applies here: https://github.com/andela/mrm_api/pull/210/files#r251803838

helpers/calendar/ratios_and_utilization.py
caveinn: I do not mean the method implementation but rather the method call.

helpers/calendar/analytics_helper.py
Bonifase: Will this renaming affect other functionalities that were using this data prior to this action?

kampkelly: No, it will not affect other functionalities. They will still work just fine. 
Though it may be better to rename the `calendar_events` later on in the places where this method is called to reflect that it is actually all events being returned.

Bonifase: If there are other functionalities that use the variable, then you would have to find out and rectify the same to avoid other parts of the application breaking due to this. Otherwise, it is just ok.

kampkelly: Okay, I will rectify this in other parts of the application by renaming them appropriately. Thanks

kampkelly: This has been done.

kampkelly:
@Bonifase the feedback has been implemented.

PR #213, opened 2019-01-29
#163391316 Allow non-admins to view analytics data

nakatuddesuzan:
#### What does this PR do?
- Allow non-admins to view analytics data
- A default user should be able to view the analytics data i.e the piecharts, graphs, and tables that are currently viewable by admins only.
- pull and checkout branch `bg-allow-non-admins-view-analytics-163391316`
- create a default user in the database 
- log into the system with that default user
- provide that user's token in the headers
- run query in the screenshot below or any other query that involves retrieving analytics data.
[#163391316](https://www.pivotaltracker.com/story/show/163391316)
#### Screenshots 
screen shot 2019-01-29 at 13 14 21

### DIFF
--- a/api/room/schema_query.py
+++ b/api/room/schema_query.py
@@ -168,7 +168,7 @@ class Query(graphene.ObjectType):
             events=resource[1]
         )
 
-    @Auth.user_roles('Admin')
+    @Auth.user_roles('Admin', 'Default User')
     def resolve_analytics_for_least_used_rooms(self, info, start_date, end_date=None):  # noqa: E501
         query = Room.get_query(info)
         room_analytics = RoomAnalytics.get_least_used_rooms_analytics(
@@ -178,7 +178,7 @@ class Query(graphene.ObjectType):
             analytics=room_analytics
         )
 
-    @Auth.user_roles('Admin')
+    @Auth.user_roles('Admin', 'Default User')
     def resolve_analytics_for_most_used_rooms(self, info, start_date, end_date=None):  # noqa: E501
         query = Room.get_query(info)
         room_analytics = RoomAnalytics.get_most_used_rooms_analytics(
@@ -189,7 +189,7 @@ class Query(graphene.ObjectType):
         )
         return room_most_used_per_week
 
-    @Auth.user_roles('Admin')
+    @Auth.user_roles('Admin', 'Default User')
     def resolve_analytics_for_meetings_per_room(self, info, start_date, end_date=None):  # noqa: E501
         query = Room.get_query(info)
         meeting_summary = RoomAnalytics.get_meetings_per_room_analytics(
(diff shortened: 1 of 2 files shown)


dnuwa:
@nakatuddesuzan this is good work though I find your commit message unclear in relation to what you accomplished. `remove all the admin decorators`

nakatuddesuzan:
Hello @dnuwa , am thanks for pointing that out. At first, I had removed the decorator but then realized I needed to use to have a better implementation. I forgot to edit the commit message afterwards Let me edit it.

api/room/schema_query.py
```diff
@@ -147,7 +147,7 @@ def resolve_room_schedule(self, info, calendar_id, days):
             events=resource[1]
         )
 
-    @Auth.user_roles('Admin')
+    @Auth.user_roles("Admin", "Default User")
```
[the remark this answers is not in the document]
Bonifase: Since we don't need admin authentication anymore to view room analytics data, I don't think admin_location_for_analytics_view method is necessary. As a matter of fact, we would need user location for analytics view if and only if necessary.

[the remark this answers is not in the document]
nakatuddesuzan: The `admin_location_for_analytics_view method` is used in other files. I also tried removing it but the tests still failed.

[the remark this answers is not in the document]
RonKbS: I'm with Bonifase on this, it the method is required then the name should be changed at least so that it reflects all users' locations.

api/user/models.py
```diff
@@ -19,7 +19,7 @@ class User(Base, Utility):
     __tablename__ = 'users'
     id = Column(Integer, Sequence('users_id_seq'), primary_key=True)
     email = Column(String, nullable=False, unique=True)
-    location = Column(String, nullable=True)
+    location = Column(String, default="Kampala")
```
dnuwa: @nakatuddesuzan kindly shade some light as to why we have a default location as Kampala

nakatuddesuzan: Okay, so if you follow up the past conversation, to do away with the decorator, we had to find a way of making sure whenever the tests run, the location of the user is saved. So this is one way that could be fixed. So if the user has no location, Kampala is saved as the location.

Bonifase: In the production environment, kindly let us know situations when a user would not have their location when setting up their Converge accounts for the first time, and why we need to make Kampala their default location hence the need to have this set in the database user model.

nakatuddesuzan: Hello @Bonifase , Here I was solving the tests problem. Following the recommendation, you gave for removing the decorator. Since the analytics view is for a specific location, the tests were failing because everytime the tests run, the location of the user is not saved in the test database. The decorator was solving this problem because the developer who designed it set a location of the user to Nairobi as a default. So if the decorator is removed, it then means, there is no way to get the user location in the test database. The other way I could implement it would be passing the location parameter in the init method in the user model class, but this would call for a change in the authentication implementation since the tokens of the user missed out the location information when they were being encoded. So by setting the default location would solve the problem. this is only affecting the test database. I chose Kampala because the offices being created while the tests run are in Kampala office.

nakatuddesuzan: The implementaion of removing the decorator works well and on insomnia, both users can view analytics because users are saved with their locations. Problem comes when tests run the user location is not saved, so the view analytics tests can not pass.

PR #216, opened 2019-01-30
#163587718 Change the names of the class and test in test-devices

RachaelNantale:
#### What does this PR do?
Change the names of the class and test in test-devices
Currently, the name of the class is `TestCreateRoom` and the test is `test-room-creation`. This though is a test for creating devices.
In case of bugs, it would be a bit difficult to know which test is failing exactly since two tests have the same name hence changing the name of the test create device
[#163587718](https://www.pivotaltracker.com/story/show/163587718)
- Not Applicable

### DIFF
--- a/tests/test_devices/test_create_device.py
+++ b/tests/test_devices/test_create_device.py
@@ -10,9 +10,9 @@ import os
 sys.path.append(os.getcwd())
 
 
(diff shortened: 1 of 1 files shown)


kampkelly:
Looks good to me

PR #220, opened 2019-01-31
#163620662 Add default value to state column

michael-basweti:
#### What does this PR do?
* Add default value to state column in all database relations that have status column.
* Added server_default parameter to the column changes in alembic revision file `e132e132ed54_add_soft_delete.py`.
* In the same file, nullable was turned to false
* git checkout to `bg-add-default-to-state-column-163620662`
* Create different resources i.e location, rooms and offices, A status field will be added in each and every relation in the database with default of `active`. When you delete anything, its status will be changed from active to archived. 
* Not applicable
[#163447376](https://www.pivotaltracker.com/n/projects/2154921/stories/163620662)


bevkololi:
> Tested out and it's good.
> Nice work @michael-basweti.
> You might need to sync with dev-ops on how to include `alembic downgrade -1` in the docker commands for migrations so that we can undo the changes.
> I have used the command to undo the changes, switch to a branch with these changes, and `alembic upgrade head` to implement the changes

+1

michael-basweti:
> Tested out and it's good.
> Nice work @michael-basweti.
> You might need to sync with dev-ops on how to include `alembic downgrade -1` in the docker commands for migrations so that we can undo the changes.
> I have used the command to undo the changes, switch to a branch with these changes, and `alembic upgrade head` to implement the changes

@RonKbS Philip has told me it is ok since I'm just editing the file, the data in staging is already active.

PR #223, opened 2019-02-01
bug(prevent-non-Andela-emails): Accept only Andela emails

dnuwa:
#### What does this PR do?
- Restricts converge users to only add Andela emails

- Validate the email provided by the user to ensure that it has the Andela domain name

- Clone the branch
- Run make build
- Run the following mutation:
````
mutation {
  createUser(email: "newuser@gmail.com"
            name: "newuser"
            picture: "www.andela.com/user"){
    user {
      email,
      name,
      picture
    }
  }
}
````
Expected response
````
{
  "errors": [
    {
      "message": "Can't add a none Andela email",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "createUser"
      ]
    }
  ],
  "data": {
    "createUser": null
  }
}
````
[#163619440](https://www.pivotaltracker.com/story/show/163619440)

#### Screenshots
image

### CODE REVIEW

File: api/user/schema.py [hunk 1]
```diff
@@ -33,6 +33,8 @@ class Arguments:
 
     def mutate(self, info, **kwargs):
         user = UserModel(**kwargs)
+        if not verify_email(user.email):
+            raise GraphQLError("Can\'t add a none Andela email")
```
cwizard2011: Please use Non-Andelan

kampkelly:
Good work @dnuwa . Please implement the requested changes and fix the codeclimate issues.

api/user/schema.py
dnuwa: @cwizard2011 this has been fixed.

nakatuddesuzan: Hey @dnuwa, great work. Since a user is created by an Admin. I feel like we can do more about this error message. It could be `You can not register a user with a non-Andelan email`. It is just a suggestion.

dnuwa: @nakatuddesuzan your message is good and very descriptive but I think its way too long. So, I think to have it short and equally descriptive is good

tests/test_user/test_create_user.py
```diff
@@ -36,3 +40,15 @@ def test_user_duplication(self):
 
         expected_response = user_duplication_mutation_response
         self.assertEqual(query_response, expected_response)
+
+    def test_user_creation_with_non_andela_email(self):
+        """
+        Testing for adding a user with non Andela Email
+        """
+        self.client.execute(non_Andela_email_mutation,
```
RonKbS: This seems to be the same code as that for line 51. Kindly share the reason for this.

dnuwa: It is unnecessary to attempt creating the same user twice, which is what the code attempts to do.  I have fixed this. @RonKbS thanx for pointing it out.

api/user/schema.py
```diff
@@ -33,6 +33,8 @@ class Arguments:
 
     def mutate(self, info, **kwargs):
         user = UserModel(**kwargs)
+        if not verify_email(user.email):
+            raise GraphQLError("Can\'t add a Non-Andelan email")
```
Bonifase: Use 'This email is not allowed' instead of 'Can't add a Non-Andelan email' just for security purposes. We don't want to suggest to intruders what email is allowed to our database.

Bonifase: Kindly check on your commit messages. Please use the guide that I shared on the slack channel last weeks

PR #228, opened 2019-02-05
 #163696871 Add firebase_token column to room model

RachaelNantale:
The push notifications feature requires a firebase_token field to be added to the Room model to identify the room to which notifications will be sent. This field can also be updated.
- checkout branch `ft-add-firebase-to-room-163696871`
- run migrations `make-migrate`
- run  the app using  the command `make run-app` 

N/A

[#163696871 ](https://www.pivotaltracker.com/story/show/163696871)


kampkelly:
Looks good to me but please fix tests and codeclimate issues.

api/room/schema.py
```diff
@@ -116,6 +116,7 @@ class Arguments:
         capacity = graphene.Int()
         image_url = graphene.String()
         calendar_id = graphene.String()
+        firebase_token = graphene.String()
```
nakatuddesuzan: Hey @RachaelNantale, Nice work, I see you included the fire_base token field in the `UpdateRoom` class enabling the field to be editable, don't we need to do the same in `CreateRoom class`. If not please throw more light?

RachaelNantale: Thanks, @nakatuddesuzan, So firstly, the story did not include the addition of the column to  the `CreateRoom class.`
Secondly, the firebase token is generated from the devices but under certain circumstances, it has to be updated. so there is no need to add it to the `CreateRoom class`. I hope this gives you more clarity. Please refer to the Android team for more information

dnuwa:
good work @RachaelNantale

PR #233, opened 2019-02-06
#163731044 Fix create response mutation

kampkelly:
#### What does this PR do?
- Allows creation of a response for multiple questions

Change question id to allow a list of integers
Modify tests accordingly

- Clone the repo.
- Run `make build`
- `make run-app`
#### Previous mutation for creating a response:
```
mutation {
  createResponse(questionId: 1, roomId: 164, rate: 1, textArea: "empty", missingItems: [65]) {
    response {
      roomId
      questionId
      question {
        question
      }
    }
  }
}
```
#### After bug fix: New mutation for creating room response

```
mutation {
  createResponse(questionId: [2, 3, 1], roomId: 164, rate: 1, textArea: "empty", missingItems: [65]) {
    response {
      roomId
      questionId
      question {
        question
      }
    }
  }
}
```

None

[#163731044](https://www.pivotaltracker.com/story/show/163731044)

#### Screenshots
Before bug fix
screen shot 2019-02-06 at 4 44 16 pm

After bug fix

_`with valid ids`_
screen shot 2019-02-07 at 3 09 56 pm

_`with some invalid ids`_
screen shot 2019-02-08 at 11 31 18 am

**Questions:**
None

### DIFF
--- a/api/response/schema.py
+++ b/api/response/schema.py
@@ -5,7 +5,8 @@ from api.response.models import Response as ResponseModel
 from utilities.validations import validate_empty_fields
 from graphql import GraphQLError
 from api.room.schema import Room
-from api.question.models import Question
+from api.question.models import Question as QuestionModel
+from api.question.schema import Question
 from helpers.response.create_response import create_response
 
 
@@ -16,13 +17,13 @@ class Response(SQLAlchemyObjectType):
 
 class CreateResponse(graphene.Mutation):
     class Arguments:
-        question_id = graphene.Int(required=True)
+        question_id = graphene.List(graphene.Int, required=True)
         room_id = graphene.Int(required=True)
         rate = graphene.Int()
         text_area = graphene.String()
         missing_items = graphene.List(graphene.Int)
 
-    response = graphene.Field(Response)
+    response = graphene.List(Response)
 
     def mutate(self, info, **kwargs):
         validate_empty_fields(**kwargs)
@@ -30,12 +31,26 @@ class CreateResponse(graphene.Mutation):
         room = query.filter_by(id=kwargs['room_id']).first()
         if not room:
             raise GraphQLError("Non-existent room id")
-        get_question = Question.query.filter_by(id=kwargs['question_id']).first() # noqa
-        if not get_question:
-            raise GraphQLError("Question does not exist")
-        question_type = get_question.question_type
-        resource = create_response(question_type, **kwargs)
-        return CreateResponse(response=resource)
+        question_ids = kwargs.pop('question_id')
+        responses = []
+        questions = Question.get_query(info).filter(
+            QuestionModel.id.in_(question_ids)).all()
+        valid_question_ids = set()
+        for question in questions:
+            valid_question_ids.add(question.id)
+            question_type = question.question_type
+            response = create_response(
+                question_type, question_id=question.id, **kwargs)
+            responses.append(response)
+        invalid_question_ids = set(question_ids).difference(
+            valid_question_ids)
+        if invalid_question_ids:
+            raise GraphQLError(
+                ('Responses for question ids {} were not saved because '
+                    'the questions do not exist').format(
+                    str(invalid_question_ids).strip('{}'))
+                )
+        return CreateResponse(response=responses)
 
 
 class Query(graphene.ObjectType):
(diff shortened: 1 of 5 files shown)


api/response/schema.py
```diff
@@ -16,26 +17,37 @@ class Meta:
...
+        # the number of question ids sent in the mutation
+        if len(questions) == len(question_ids):
+            for question in questions:
+                question_type = question.question_type
+                resource = create_response(
+                    question_type, question_id=question.id, **kwargs)
+                resources.append(resource)
+        else:
+            raise GraphQLError(
+                "One or more of the provided questions does not exist"
```
bevkololi: @kampkelly, slight concern, if you are checking if the questions that the user has input exist, is checking that using the length a good idea? The lengths could be similar but the question ids also inexistent. For example existing questions are [1,2,3] and I input [3,4,5], length is still the same but they don't exist. From what I'm seeing this scenario doesn't raise this error. Kindly correct me if I'm wrong.

dnuwa: looking at @bevkololi 's concern I think the use of sets could make it better. i.e `question_ids=[1,2,3,4,5,6,7,8,9,0]` and `question_id =[4,5,6]`
then:
`set(question_ids)&set(question_id)` returns `[4,5,6]`

caveinn: I agree with @dnuwa on this.

bevkololi: True @dnuwa. Or simply looping through the argument and checking if it exists in the question_id list.

caveinn: I also have another concern. When validating the data passed in, is there a way to validate that for each question passed,  the response is of the type the question expects. For instance if I pass two questions that expect a `rate` type and pass only one rate type it will give both questions that same rating.

kampkelly: @caveinn Thanks for the feedback. I see your concern but that may have to be handled in another story. My task is to allow saving multiple questions at a time and not to validate the question_type.

bevkololi: These validations already existed @kampkelly, so it'll just be finding a way to incorporate them to this new structure of response mutations.

kampkelly: @bevkololi  and @dnuwa thanks for your feedback. But I belive checking with the length will work as expected. This is because the code on line 36 in api/response/schema.py checks if there are questions with the id's sent in the mutation and returns a list of this questions if found.
So if you pass id's [3,4,5] of which two id's do not exist. It returns a list of count 1.
So when I check if the length is the same with the input sent (count 3). This will not be equal and so the error will be raised. 
I can explain this to everyone during the backend sync today.

Bonifase: I would suggest  that you be more specific with your error messages.

kampkelly: Thanks @Bonifase I will work on that.

kampkelly: This has been fixed now.

api/response/schema.py
```diff
@@ -16,26 +17,39 @@ class Meta:
...
+            valid_question_ids.add(question.id)
+            question_type = question.question_type
+            resource = create_response(
+                question_type, question_id=question.id, **kwargs)
+            resources.append(resource)
+        invalid_question_ids = set(question_ids).difference(
+            valid_question_ids)
+        if invalid_question_ids:
+            raise GraphQLError(
+                'Response for question ids {} were not saved'.format(
```
Bonifase: Tell the user why they were not saved. Also, write the test for this.

kampkelly: This has been implemented @Bonifase.

api/response/schema.py
```diff
@@ -16,26 +17,40 @@ class Meta:
...
         if not room:
             raise GraphQLError("Non-existent room id")
-        get_question = Question.query.filter_by(id=kwargs['question_id']).first() # noqa
-        if not get_question:
-            raise GraphQLError("Question does not exist")
-        question_type = get_question.question_type
-        resource = create_response(question_type, **kwargs)
-        return CreateResponse(response=resource)
+        question_ids = kwargs.pop('question_id')
+        resources = []
```
RonKbS: From what is going on in the code, I would think this should be `response` and not `resource`. Same for line 42. Kindly share your thoughts on this, @kampkelly.

kampkelly: Thanks @RonKbS for the feedback. 
This has been renamed to response and other places accordingly.

dnuwa:
good work @kampkelly. the functionality works fine.

PR #234, opened 2019-02-06
#163653690 Fix cancellation percentage accuracy

RonKbS:
#### What does this PR do?
The PR increases the number of decimal places of the cancellation percentage.
It also adjusts the function used to return the percentage to make it a member of the class in which it had been placed.
In regards to the assertion added in the associated tests file, the line that returns the percentage is not covered and the assertion resolves this.

Enable a more accurate figure to be returned when running analytics rations, more especially cancellationsPercentage.

Pull the branch and create an odd number of events with a valid calender_id corresponding to a room_id in the local database.
Run the mutation below with corresponding dates for the events created.

``` 
query {
  analyticsRatios (startDate: "Jul 05 2018", endDate: "Jul 16 2018") {
    cancellationsPercentage
    cancellations
    appBookings
    appBookingsPercentage
  }
}
```

N/A

[#163653690](https://www.pivotaltracker.com/story/show/163653690)

Before
screen shot 2019-02-01 at 2 42 46 pm

After making changes
screenshot 2019-02-07 at 20 40 42

### DIFF
--- a/helpers/calendar/ratios_and_utilization.py
+++ b/helpers/calendar/ratios_and_utilization.py
@@ -128,13 +128,13 @@ class RoomAnalyticsRatios(Credentials):
         bookings = len(calendar_events_list) + len(
             app_events_list) + cancellations
         app_bookings = len(app_events_list) + cancellations
-        app_bookings_percentage = RoomAnalyticsRatios.percentage_formater(
+        app_bookings_percentage = RoomAnalyticsRatios().percentage_formater(
             app_bookings,
             bookings)
-        cancellations_percentage = RoomAnalyticsRatios.percentage_formater(
+        cancellations_percentage = RoomAnalyticsRatios().percentage_formater(
             cancellations,
             bookings)
-        checkins_percentage = RoomAnalyticsRatios.percentage_formater(
+        checkins_percentage = RoomAnalyticsRatios().percentage_formater(
             checkins,
             bookings)
         result = RatioOfCheckinsAndCancellations(
@@ -149,14 +149,14 @@ class RoomAnalyticsRatios(Credentials):
             )
         return result
 
-    def percentage_formater(portion, total):
+    def percentage_formater(self, portion, total):
         """ Calculates the percentage of the entered portion to the total and returns it
          :params
             - portion, total
         """
         try:
             percentage = (portion/total) * 100
-            return round(percentage, 1)
+            return percentage
         except ZeroDivisionError:
             return 0
 
--- a/fixtures/events/events_ratios_fixtures.py
+++ b/fixtures/events/events_ratios_fixtures.py
@@ -70,3 +70,18 @@ event_ratio_per_room_response = {
         }
     }
 }
+
+event_ratio_percentage_cancellation_query = '''query {
+    analyticsRatios(startDate:"Jul 10 2018", endDate:"Jul 29 2018"){
+        cancellationsPercentage
+}
+}
+'''
+
+event_ratio_percentage_cancellation_response = {
+    "data": {
+        "analyticsRatios": {
+            "cancellationsPercentage": 100.0
+        }
+    }
+}
--- a/tests/test_events/test_event_checkin.py
+++ b/tests/test_events/test_event_checkin.py
@@ -8,6 +8,9 @@ from fixtures.events.event_checkin_fixtures import (
     cancel_event_mutation,
     cancel_event_respone
 )
+from fixtures.events.events_ratios_fixtures import (
+    event_ratio_percentage_cancellation_query,
+    event_ratio_percentage_cancellation_response)
 
 sys.path.append(os.getcwd())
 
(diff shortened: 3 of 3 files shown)


helpers/calendar/ratios_and_utilization.py
```diff
@@ -149,14 +149,14 @@ def map_results_to_ratio_class(self,
...
-    def percentage_formater(portion, total):
+    def percentage_formater(self, portion, total):
         """ Calculates the percentage of the entered portion to the total and returns it
          :params
             - portion, total
         """
         try:
             percentage = (portion/total) * 100
-            return round(percentage, 1)
+            return round(percentage, 2)
```
nakatuddesuzan: Hey @RonKbS, Nice work, I have a small concern, the task is requiring us to be as accurate as possible, Is rounding off to two decimal places the best we can go for or we can go ahead and round off to 3 or 4. Or you could include clear screenshots that show the differences between those two results other than showing the difference in the decimal places. This will show more about the accuracy of 2 decimal places.

kampkelly:
Good work @RonKbS but please look into the concern raised by @nakatuddesuzan.
The percentage is meant to be as accurate as possible.

helpers/calendar/ratios_and_utilization.py
RonKbS: I have gone with removing the removing the round function as recommended by the TC @nakatuddesuzan , @kampkelly and @caveinn.

Tests are failing now because I have included a check for calender_id to make sure data is not corrupted by creating rooms with similar calender_ids. This is because the id is used in analytics to retrieve data for rooms

bevkololi: Hey @RonKbS, isn't this out of scope?

RonKbS: I'm seeing it as being part of 'increasing the accuracy of the cancellation percentage'.

The story's description also refers to bookings. I, therefore, thought to add the calender_id check as another means of improving the accuracy because in my case, for example, the percentages returned on running the analytics were similar. This was because the rooms I created had the same calender_id.

api/room/schema.py
```diff
@@ -90,9 +90,21 @@ def mutate(self, info, office_id, **kwargs):
...
         query = Room.get_query(info)
         exact_query = room_join_office(query)
+
+        # prevent analytics data being corrupted by
+        # checking that rooms created have unique calendar ids
+        room_with_same_calender_id = query.filter(
+            RoomModel.state == "active",
+            RoomModel.calendar_id == kwargs['calendar_id']
+        )
+        if room_with_same_calender_id.count() > 0:
```
Bonifase: Why do we have to count, can't we just check without counting?

fixtures/room/create_room_fixtures.py
```diff
@@ -164,7 +164,7 @@
     mutation {
         createRoom(
             name: "Kigali", roomType: "Meeting", capacity: 6, floorId: 4, officeId: 1,
-            calendarId:"andela.com_38363233383232303439@resource.calendar.google.com",
+            calendarId:"232k",
```
Bonifase: Why was this changed?

Bonifase: I would suggest not to round off the percentage values, let the front end team decide on the number of decimal places they'd like to render.

api/room/schema.py
RonKbS: An object is returned whether it is a positive or negative result as regards the if-check.
.count is, however, able to obtain a negative result on the query result that is run in this case and hence it is employed here.

fixtures/room/create_room_fixtures.py
RonKbS: The test that used this fixture was hitting the check for existing calender_ids and stopping at that point, returning the error raised by that check.
This change ensures it moves past it and does what is required by the test.

helpers/calendar/ratios_and_utilization.py
```diff
@@ -149,14 +149,14 @@ def map_results_to_ratio_class(self,
...
-    def percentage_formater(portion, total):
+    def percentage_formater(self, portion, total):
         """ Calculates the percentage of the entered portion to the total and returns it
          :params
             - portion, total
         """
         try:
             percentage = (portion/total) * 100
-            return round(percentage, 1)
+            return percentage
```
RonKbS: > I would suggest not to round off the percentage values as shown in the screenshots, let the front end team decide on the number of decimal places they'd like to render.

The rounding off has been removed but the snapshot has not been updated @Bonifase.
Let me update it now.

dnuwa:
good work @RonKbS I've tested it out and it runs fine.

RonKbS:
> This is quality work @RonKbS, kudos! Please however raise a separate PR for the analytics fix. We'll losing valuable commit history if I merge this change a long with this PR.

Thank you @joshuaocero.

@dnuwa was recently given a task that also implements this `https://github.com/andela/mrm_api/pull/246/files` so I will go ahead and remove it from the PR.

PR #235, opened 2019-02-07
#163741006 Fix edit and delete location

RachaelNantale:
#### What does this PR do?
The PR ensures that the admins should not be able to edit or delete locations that are not their own.

Currently, an admin can edit or delete a location where they do no reside. For example, an Admin in `Kampala` can update the location of `Nairobi`. This PR ensures that admins can only make changes to the location that they belong to.

- Pull the branch `bg-fix-edit-delete-location-163741006`
- run the command `make run-app`
- Run the `createLocation` query with Nairobi as the new location.
- Run the following query

``` 
mutation{
    deleteLocation(locationId:1){
        location{
        id
        }
    }
    }
```
**OR**
```
mutation{
    updateLocation(locationId: 1, name: "Nairobi", country: "Kenya", abbreviation: "KE"){ # noqa: E501
        location{
            name
            abbreviation
            country
        }
    }
}
```

An admin  can create any location but is not authorized to make changes to it if he/she is not in that specific location

[#163741006](https://www.pivotaltracker.com/story/show/163741006)

Before **updateLocation**
screen shot 2019-02-07 at 9 48 29 am

After making changes
screen shot 2019-02-07 at 9 49 12 am

Before **deleteLocation**
screen shot 2019-02-07 at 9 49 54 am

After making changes
screen shot 2019-02-07 at 9 50 23 am

### DIFF
--- a/helpers/auth/admin_roles.py
+++ b/helpers/auth/admin_roles.py
@@ -19,6 +19,14 @@ class Admin_roles():
         if admin_details.location != location.name:
             raise GraphQLError("You are not authorized to make changes in " + location.name)  # noqa: E501
 
+    def verify_admin_location(self, location_id):
+        admin_details = get_user_from_db()
+        location = Location.query.filter_by(id=location_id).first()
+        if admin_details.location != location.name:
+            raise GraphQLError(
+                "You are not authorized to make changes in " + location.name
+            )
+
     def update_delete_rooms_create_resource(self, room_id):
         admin_details = get_user_from_db()
         location_query = location_join_room()
@@ -39,12 +47,6 @@ class Admin_roles():
         floor_location = location_query.filter_by(id=floor_id).first()
         if admin_details.location != floor_location.name:
             raise GraphQLError("You are not authorized to make changes in " + floor_location.name)  # noqa: E501
-
-    def create_office(self, location_id):
-        admin_details = get_user_from_db()
-        location = Location.query.filter_by(id=location_id).first()
-        if location.name != admin_details.location:
-            raise GraphQLError("You are not authorized to make changes in " + location.name)  # noqa: E501
 
     def check_office_location_create_wing(self, floor_id):
         get_floor = FloorModel.query.filter_by(id=floor_id).first()
--- a/api/location/schema.py
+++ b/api/location/schema.py
@@ -17,6 +17,7 @@ from utilities.utility import update_entity_fields
 from helpers.room_filter.room_filter import room_join_location
 from helpers.auth.validator import ErrorHandler
 from helpers.auth.authentication import Auth
+from helpers.auth.admin_roles import admin_roles
 
 
 class Location(SQLAlchemyObjectType):
@@ -67,6 +68,7 @@ class UpdateLocation(graphene.Mutation):
             LocationModel.id == location_id).first()
         if not location_object:
             raise GraphQLError("Location not found")
+        admin_roles.verify_admin_location(location_id)
         if "time_zone" in kwargs:
             validate_timezone_field(**kwargs)
         if "country" in kwargs:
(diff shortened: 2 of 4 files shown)


helpers/auth/admin_roles.py
```diff
@@ -40,12 +48,6 @@ def update_delete_floor(self, floor_id):
         if admin_details.location != floor_location.name:
             raise GraphQLError("You are not authorized to make changes in " + floor_location.name)  # noqa: E501
 
-    def create_office(self, location_id):
```
RachaelNantale: I removed this because it was a duplicate function and used one with a better descriptive name

kampkelly:
Please fix codeclimate issues.

helpers/auth/admin_roles.py
```diff
@@ -19,6 +19,14 @@ def create_rooms_update_delete_office(self, office_id):
         if admin_details.location != location.name:
             raise GraphQLError("You are not authorized to make changes in " + location.name)  # noqa: E501
 
+    def check_location(self, location_id):
```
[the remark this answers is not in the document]
michael-basweti: > Nice work @RachaelNantale , from what the function does I believe something like `verify_admin_location` would be more descriptive. What do you think?

I agree with @C3real-kill3r

helpers/auth/admin_roles.py
```diff
@@ -19,6 +19,14 @@ def create_rooms_update_delete_office(self, office_id):
...
+        admin_details = get_user_from_db()
+        location = Location.query.filter_by(id=location_id).first()
```
caveinn: I think for this you filter you should also add state=active to ensure you do not capture locations that have been archived.

helpers/auth/admin_roles.py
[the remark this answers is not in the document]
RachaelNantale: okay lemme work on this.

helpers/auth/admin_roles.py
RachaelNantale: Thanks, @caveinn, this is being implemented in the different classes. I also had brought it up but @Bonifase emphasised during the sync that we need to prioritise the other tasks and since its implementation is okay. We can always create another story for refactoring code.

dnuwa:
this runs fine an meets the story description needs.

helpers/auth/admin_roles.py
```diff
@@ -19,6 +19,14 @@ def create_rooms_update_delete_office(self, office_id):
         if admin_details.location != location.name:
             raise GraphQLError("You are not authorized to make changes in " + location.name)  # noqa: E501
 
+    def verify_admin_location(self, location_id):
+        admin_details = get_user_from_db()
+        location = Location.query.filter_by(id=location_id).first()
+        if admin_details.location != location.name:
```
joshuaocero: @RachaelNantale considering that when adding a user, location is added as a string, how can we guarantee that the user location string and location name will always match?

RachaelNantale: Thank you @joshuaocero, So after the brief sync we had I went through the code and discovered that there are two ways of creating a user. 
The Auto creation when one is interacting with the other queries that need a user and the manual creation using the fixture.
When a user is created automatically, the email, name and picture are saved. The location, on the other hand, is gotten from the `andela api` when a request is sent using the email from the email and token.
So the location is not created by us.

joshuaocero: Noted @RachaelNantale. True that the user location is not created by us but the location name on the Location model is. Kindly add a TODO comment so we remember to refactor this section when we reorganise the User <> Location relationship.

RachaelNantale: Okay @joshuaocero, this has been implemented.

PR #239, opened 2019-02-07
#163729499 Fix analytics date range for bookingsAnalyticsCount query

caveinn:
#### What does this PR do?
- It enables the user to enter a date range of their choice 

- Remove the restriction to sending a date range that is bellow 15 days and greater than
90 days. This will enable the user to choose the range freely.

- Pull this branch.
- Run the app using `make run-app`
- run the query below.
```
query{
  bookingsAnalyticsCount(startDate:"jan 2 2019" endDate: "jan 25 2019"){
    bookings
    period
}
}
```

N/A

[#163729499](https://www.pivotaltracker.com/n/projects/2154921/stories/163729499)

The query response before the fix
image

The query response after the fix
image

The query response if the period is greater than 31 days
image

#### Questions:

### DIFF
--- a/helpers/calendar/ratios_and_utilization.py
+++ b/helpers/calendar/ratios_and_utilization.py
@@ -1,5 +1,4 @@
 import dateutil.parser
-from graphql.error import GraphQLError
 
 from .credentials import Credentials
 from helpers.calendar.analytics_helper import (CommonAnalytics)
@@ -167,7 +166,7 @@ class RoomAnalyticsRatios(Credentials):
         end_dt = dateutil.parser.parse(day_after_end_date)
         number_of_days = (end_dt - start_dt).days
 
-        if number_of_days <= 15:
+        if number_of_days <= 30:
             dates = CommonAnalytics.get_list_of_dates(start, number_of_days)  # noqa E501
             for date in dates:
                 bookings = CommonAnalytics.get_total_bookings(self, query, date[0], date[1])  # noqa E501
@@ -175,15 +174,12 @@ class RoomAnalyticsRatios(Credentials):
                 output = BookingsAnalyticsCount(period=string_date, bookings=bookings) # noqa E501
                 results.append(output)
 
-        elif number_of_days >= 90:
+        else:
             dates = CommonAnalytics.get_list_of_month_dates(start_date, start_dt, day_after_end_date, end_dt)  # noqa E501
             for date in dates:
                 bookings = CommonAnalytics.get_total_bookings(self, query, date[0], date[1])  # noqa E501
                 string_month = dateutil.parser.parse(date[0]).strftime("%B")
                 output = BookingsAnalyticsCount(period=string_month, bookings=bookings) # noqa E501
                 results.append(output)
-
-        else:
-            raise GraphQLError("Kindly enter a valid date range(less than 15 days or greater than 90 days")  # noqa E501
 
         return results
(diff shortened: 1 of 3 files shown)


tests/test_rooms/test_room_analytics.py
```diff
@@ -108,20 +102,3 @@ def test_analytics_for_daily_bookings(self):
             self, get_bookings_count_daily,
             get_bookings_count_daily_response
         )
-
-    def test_analytics_for_monthly_bookings(self):
```
nakatuddesuzan: Hello @caveinn , Kindly throw more light on why this test was removed

tests/test_rooms/test_room_analytics.py
```diff
@@ -108,20 +102,3 @@ def test_analytics_for_daily_bookings(self):
...
-            get_bookings_count_monthly_response
-        )
-
-    def test_analytics_for_bookings_invalid_date_range(self):
-        CommonTestCases.admin_token_assert_equal(
-            self, get_bookings_count_invalid_date_range,
-            get_bookings_count_invalid_date_range_response
-        )
-
-    def test_analytics_for_monthly_bookings_diff_years(self):
```
nakatuddesuzan: This one too

kampkelly:
The code looks good, please work on the test coverage @caveinn

helpers/calendar/ratios_and_utilization.py
```diff
@@ -167,23 +166,20 @@ def get_bookings_analytics_count(self, query, start, end):
...
                 results.append(output)
 
-        elif number_of_days >= 90:
+        else:
             dates = CommonAnalytics.get_list_of_month_dates(start_date, start_dt, day_after_end_date, end_dt)  # noqa E501
             for date in dates:
                 bookings = CommonAnalytics.get_total_bookings(self, query, date[0], date[1])  # noqa E501
                 string_month = dateutil.parser.parse(date[0]).strftime("%B")
                 output = BookingsAnalyticsCount(period=string_month, bookings=bookings) # noqa E501
                 results.append(output)
```
michael-basweti: I feel like since we are not limiting a user to a specific date, we should just have one function i.e a function to show us the analytics for any given date. What I mean is we should maybe not care if the data we get is monthly or for 3 days so I think we should do away with the `if statement`. What do you think?

caveinn: This is because of the date range is greater than a month the data should not be returned as daily data but rather monthly data. This will ensure the fronted have data that they can easily display.

tests/test_rooms/test_room_analytics.py
```diff
@@ -115,12 +113,6 @@ def test_analytics_for_monthly_bookings(self):
             get_bookings_count_monthly_response
         )
 
-    def test_analytics_for_bookings_invalid_date_range(self):
-        CommonTestCases.admin_token_assert_equal(
-            self, get_bookings_count_invalid_date_range,
-            get_bookings_count_invalid_date_range_response
-        )
-
     def test_analytics_for_monthly_bookings_diff_years(self):
```
caveinn: This test was removed because it was testing the invalid dates but with the current implementation there are no invalid dates.

dnuwa:
tested and runs as required. good work @caveinn

helpers/calendar/ratios_and_utilization.py
michael-basweti: > This is because of the date range is greater than a month the data should not be returned as daily data but rather monthly data. This will ensure the fronted have data that they can easily display.

Awesome clarification

PR #243, opened 2019-02-11
#161947308 Implement cancelation of unchecked-in events

C3real-kill3r:
#### What does this PR do?
- [x] add query to get all recurring events per room
- [x] Preload event model with google api event data
- [x] Create check-in relationship
- [x] Implement functionality to cancel event after 3 consecutive missed checkins

- A recurring meeting should be cancelled when a user fails to check in consecutively 3 times. This is to avoid ghost meetings and make the room available for use.
- The implementation includes a file that can be run to check for events that have not been checked in for 3 consecutive times.

- Pull this branch
- make the migrations as explained in the README
- create a room in your database with the following calendar_id `andela.com_3131313630383336383038@resource.calendar.google.com`
- create a recurring event and add converge test room, preferably over the weekends
- make changes in the `update function` as shown below from line 91 to 96, and manually add the date in which your recurring event falls (on line 91).

screenshot 2019-02-09 at 09 05 05

- on the same file, comment out line 138 to 143 like below,

screenshot 2019-02-09 at 12 23 54

- to simulate three missed checkins, run  `python run_autocancel_room.py` 3 concecutive times.
- the events model will be populated three times with the recurring events for the specified dates and state will toggle to `archived` and the respective event will be cancelled in the calendar.

#### Below are the screenshots of a canceled room event
image

- The data relating to this task is gotten from the google calendar API.
- The file is included outside the API since it cannot be auto tested.
- When testing this ensure you have the same environment variables as you would when running the API.
- The file should be run automatically every day to clear all events that are recurring and have not been checked into it.

[#161947308](https://www.pivotaltracker.com/story/show/161947308)


services/room_cancelation/auto_cancel_event.py
```diff
@@ -0,0 +1,133 @@
...
+                        "date_of_event": start_date,
+                        "event_id": event.get("id"),
+                        "recurring_event_id": recurring_event_id,
+                        "room_id": room.id,
+                        "calendar_id": room.calendar_id
+                    }
+                    recurring_events.append(recurring_event)
+        return recurring_events
+
+    def update(self):
```
Bonifase: Use a descriptive function name, what are we updating here?

C3real-kill3r: we're updating the event status, @Bonifase

PR #244, opened 2019-02-11
#163783849 Add missing responses count for a particular question and room.

RonKbS:
#### What does this PR do?
The PR enables the retrieval of the total responses for a question in a particular room.

Enable the query that uses question_id to retrieve the number of responses per room.

Pull the branch and run the following query.

``` 
query {
    question(id: 1){
        question
        id
        response {
          room {
            id
            name
          }
    	  questionResponseCountInRoom
          rate
          check
        }
    }
}
```

N/A

[#163783849](https://www.pivotaltracker.com/story/show/163783849)

screenshot 2019-02-12 at 08 32 20

### DIFF
--- a/api/response/models.py
+++ b/api/response/models.py
@@ -34,3 +34,9 @@ class Response(Base, Utility):
         secondary="missing_items",
         backref=('resources'),
         lazy="joined")
+
+    @property
+    def question_response_count_in_room(self):
+        return Response.query.filter_by(
+            question_id=self.id, room_id=self.room_id
+        ).count()
--- a/api/response/schema.py
+++ b/api/response/schema.py
@@ -13,6 +13,8 @@ class Response(SQLAlchemyObjectType):
     class Meta:
         model = ResponseModel
 
+    question_response_count_in_room = graphene.Int()
+
 
 class CreateResponse(graphene.Mutation):
     class Arguments:
--- a/fixtures/questions/get_question_fixtures.py
+++ b/fixtures/questions/get_question_fixtures.py
@@ -142,6 +142,7 @@ query {
     endDate
     questionResponseCount
     response {
+      questionResponseCountInRoom
       id
       questionId
       roomId
@@ -160,6 +161,7 @@ get_question_by_id_query_response = {
             'endDate': '28 Nov 2018',
             'questionResponseCount': 1,
             'response': [{
+                'questionResponseCountInRoom': 1,
                 'id': '1',
                 'questionId': 1,
                 'roomId': 1


api/response/schema.py
```diff
@@ -13,6 +13,8 @@ class Response(SQLAlchemyObjectType):
     class Meta:
         model = ResponseModel
 
+    room_response_count = graphene.Int()
```
Bonifase: is it room response or question response?

RonKbS: There is already a question_response variable in the query as can be shown in the screenshot.

api/response/models.py
```diff
@@ -34,3 +34,9 @@ class Response(Base, Utility):
         secondary="missing_items",
         backref=('resources'),
         lazy="joined")
+
+    @property
+    def room_response_count(self):
+        return Response.query.filter_by(
+            question_id=self.id, room_id=self.room_id
+        ).count()
```
caveinn: @RonKbS , Will this include response count to rooms that have been soft-deleted?

RonKbS: @caveinn, when a room is deleted, any related tables will have the associated data removed. This is the case for hard-delete however. I think the implementation of soft delete is incomplete in this case, therefore. This is because data in associated tables is not affected by soft delete. As an example, you can take a look at the question's delete mutation. It does not account for the responses associated with a particular question.

So yes, as of now this includes responses count for archived rooms.

As a way forward a story may need to be raised that prevents queries that pick up data from related tables from including archived data as well.

api/response/schema.py
RonKbS: I have updated it to question_response_count_in_room, @Bonifase.

api/response/models.py
```diff
@@ -34,3 +34,9 @@ class Response(Base, Utility):
         secondary="missing_items",
         backref=('resources'),
         lazy="joined")
+
```
michael-basweti: Hey @RonKbS , I think you should calculate the responses in the room query in the schema instead of having this property in the models. If done in the schema it will reduce the boilerplate.

api/response/models.py
bevkololi: +1 to what @caveinn said, I think you should only query the Responses that have not been archived, by simply changing the way you filter it. This has been done in prior stories too.

api/response/models.py
bevkololi: Also, I strongly feel that this could have been approached best by refactoring `question/schema_query.py`. There's part of this functionality there, the problem is that the query did not take a question id yet. Else we could just do away with the whole file because it is not used at all. Thoughts?

RonKbS: Doing this in the schema will not make the data available to other queries because I am taking advantage of relations to access the responses from the questions. Also, I do not think this is (boilerplate)[https://en.wikipedia.org/wiki/Boilerplate_code] because the code is not similar to other pieces in the codebase.

api/response/models.py
RonKbS: The responses do not have a state column, @bevkololi.

api/response/models.py
bevkololi: `Else we could just do away with the whole file because it is not used at all. Thoughts?` ?

RonKbS: @bevkololi the refactoring of that file would require much more time and this PR has been raised to unblock an urgent task in the frontend. Welike Amos can be contacted for more information on this.

api/response/models.py
kampkelly: So, do we need a story for this (responses to have a state column)?

dnuwa:
Good work @RonKbS. I have tested it on and it works fine.

joshuaocero:
@RonKbS between these two endpoints; `getRoomResponse` and `allRoomResponses`, aren't we already able to return the required data?

RonKbS:
> @RonKbS between these two endpoints; `getRoomResponse` and `allRoomResponses`, aren't we already able to return the required data?

The data returned by those two queries is not refined according to question id. In addition to that, the particular query worked on this PR is the one the frontend team requested changes for.

joshuaocero:
This feature may no longer be a requirement. I am following it up further with the product team.

PR #245, opened 2019-02-11
#163782051 get remote rooms from google calendar api

kampkelly:
#### What does this PR do?
- Gets the list of rooms from google calendar api

Return all rooms from the google calendar api

- Clone the repo.
- Run `make build`
- `make run-app`

#### Run the following query:
```
query {
   allRemoteRooms {
    rooms {
      calendarId
      name
    }
  }
}
```

None

[#163782051](https://www.pivotaltracker.com/story/show/163782051)

#### Screenshots
screen shot 2019-02-11 at 4 31 09 pm

**Questions:**
None

### DIFF
--- a/api/room/schema_query.py
+++ b/api/room/schema_query.py
@@ -11,6 +11,7 @@ from api.room.schema import (RatioOfCheckinsAndCancellations,
                              BookingsAnalyticsCount)
 from helpers.pagination.paginate import ListPaginate
 from helpers.calendar.analytics_helper import CommonAnalytics
+from helpers.calendar.credentials import Credentials
 
 
 class Analytics(graphene.ObjectType):
@@ -23,6 +24,15 @@ class Analytics(graphene.ObjectType):
 
 class RatiosPerRoom(graphene.ObjectType):
     ratios = graphene.List(RatioOfCheckinsAndCancellations)
+
+
+class RemoteRoom(graphene.ObjectType):
+    calendar_id = graphene.String()
+    name = graphene.String()
+
+
+class AllRemoteRooms(graphene.ObjectType):
+    rooms = graphene.List(RemoteRoom)
 
 
 class CalendarEvent(graphene.ObjectType):
@@ -53,6 +63,8 @@ class Query(graphene.ObjectType):
         Room,
         room_id=graphene.Int()
     )
+
+    all_remote_rooms = graphene.Field(AllRemoteRooms)
 
     get_room_by_name = graphene.List(
         Room,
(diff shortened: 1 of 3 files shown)


api/room/schema_query.py
```diff
@@ -140,6 +152,25 @@ def resolve_all_rooms(self, info, **kwargs):
         response = PaginatedRooms(**kwargs)
         return response
 
+    def resolve_all_remote_rooms(self, info):
```
michael-basweti: According to your tests, seems only the admin should be able to get the list of rooms so maybe you should include `@Auth.user_roles('Admin')` before your function. Unless anybody is allowed to access in which case maybe the test should be refactored

kampkelly: Thanks @michael-basweti for the feedback, I just asked @joshuaocero and he said only admins should be able to do this, so I will add the decorator to the method.

michael-basweti: > Thanks @michael-basweti for the feedback, I just asked @joshuaocero and he said only admins should be able to do this, so I will add the decorator to the method.
@kampkelly , that will work

kampkelly: Great, it has been done.

caveinn:
image
I think you should validate the calendar_ids (I advice to use regex) to avoid some wrong ids as seen above.

C3real-kill3r:
> image
> 
> I think you should validate the calendar_ids (I advice to use regex) to avoid some wrong ids as seen above.

@kampkelly you can split the id from(_)...then check if it the first element has "andela.com". Otherwise it's working well though as pointed above it returns some wrong data

kampkelly:
> image
> 
> I think you should validate the calendar_ids (I advice to use regex) to avoid some wrong ids as seen above.

Thanks @caveinn for the feedback, I had shown @joshuaocero these ids which seem different from the others. His recommendation was that I return all the calendar_ids.

kampkelly:
> > image
> > I think you should validate the calendar_ids (I advice to use regex) to avoid some wrong ids as seen above.
> 
> @kampkelly you can split the id from(_)...then check if it the first element has "andela.com". Otherwise it's working well though as pointed above it returns some wrong data

please see my response to @caveinn.
However, I will consult with @joshuaocero again on this.

kampkelly:
@caveinn and @C3real-kill3r I have discussed with @joshuaocero and I have now done a filter to get only those that have valid room ids.

api/room/schema_query.py
```diff
@@ -25,6 +26,15 @@ class RatiosPerRoom(graphene.ObjectType):
     ratios = graphene.List(RatioOfCheckinsAndCancellations)
 
 
+class RemoteRooms(graphene.ObjectType):
```
RonKbS: Great work @kampkelly.

My suggestion is to use RemoteRoom here since you're dealing with one object of sorts.

dnuwa:
Good Work @kampkelly I've tested it out and it works as expected. kindly look into @RonKbS concern.

api/room/schema_query.py
kampkelly: Thanks @RonKbS . This has been done.

kampkelly:
> Fix the failing tests

Thanks @Bonifase . I am working on it.

api/room/schema_query.py
```diff
@@ -152,6 +164,29 @@ def resolve_all_rooms(self, info, **kwargs):
...
 
+    @Auth.user_roles('Admin')
+    def resolve_all_remote_rooms(self, info):
+        service = Credentials().set_api_credentials()
+        page_token = None
+        remote_rooms = []
+        while True:
+            calendar_list = service.calendarList().list(
+                pageToken=page_token).execute()
+            for calendar_list_entry in calendar_list['items']:
```
Bonifase: Consider using room_object instead of calendar_list_entry. Kindly remove the \ in line 176

kampkelly: Thanks @Bonifase. This has been implemented.

PR #247, opened 2019-02-11
#163849417 Refactor event check-in endpoint

C3real-kill3r:
#### What does this PR do?
- [x] add an helper function to check the status of events
- [x] refactor check-in endpoint

Make the checkin feature more robust by:

 - Only updating `checkin field` to `True` if the event exists in the DB
 - Add the event to the DB if it does not exist with checkin set to `True`

- Pull this branch
- make the migrations as explained in the [README](https://github.com/andela/mrm_api/blob/develop/Readme.md)
- create an event and run the following query to check-in into that event

```
mutation {
  eventCheckin(calendarId: "your_calendar_id", eventId: "the_event_id", eventTitle: "event_title", startTime: "2019-02-15T01:00:00-08:00", endTime: "2019-02-15T01:30:00-08:00") {
    event {
      eventId
      roomId
      checkedIn
      cancelled
      room {
        id
        name
        calendarId
      }
    }
  }
}

```

### Below are the screenshots 
* successful check-in into an event
screenshot 2019-02-11 at 22 22 56

* trying to check-in into an already checked in event
screenshot 2019-02-11 at 22 24 15

* checking in into a canceled event
screenshot 2019-02-11 at 22 25 27

N/A

[#163849417](https://www.pivotaltracker.com/story/show/163849417)

### DIFF
--- a/api/events/schema.py
+++ b/api/events/schema.py
@@ -1,28 +1,13 @@
 import graphene
 from graphene_sqlalchemy import SQLAlchemyObjectType
-from graphql import GraphQLError
 
 from api.events.models import Events as EventsModel
-from api.room.models import Room as RoomModel
+from helpers.calendar.events import RoomSchedules
 
 
 class Events(SQLAlchemyObjectType):
     class Meta:
         model = EventsModel
-
-    def check_status(self, info, **kwargs):
-        try:
-            room_id = RoomModel.query.filter_by(
-                calendar_id=kwargs['calendar_id']).first().id
-            if EventsModel.query.filter_by(
-                    event_id=kwargs['event_id'],
-                    room_id=room_id).count() > 0:
-                raise GraphQLError("You cannot perform this action")
-            return room_id
-
-        except AttributeError:
-            raise GraphQLError(
-                "This Calendar ID is not registered on Converge.")
 
 
 class EventCheckin(graphene.Mutation):
@@ -35,18 +20,24 @@ class EventCheckin(graphene.Mutation):
     event = graphene.Field(Events)
 
     def mutate(self, info, **kwargs):
-        room_id = Events.check_status(self, info, **kwargs)
-        new_event = EventsModel(
-            event_id=kwargs['event_id'],
-            room_id=room_id,
-            event_title=kwargs['event_title'],
+        room_id = RoomSchedules().check_event_status(info, **kwargs)
+        event = EventsModel.query.filter_by(
             start_time=kwargs['start_time'],
-            end_time=kwargs['end_time'],
-            checked_in=True,
-            cancelled=False)
-        new_event.save()
-
-        return EventCheckin(event=new_event)
+            event_id=kwargs['event_id']).scalar()
+        if event is not None:
+            event.checked_in = True
+            event.save()
+        else:
+            event = EventsModel(
+                event_id=kwargs['event_id'],
+                room_id=room_id,
+                event_title=kwargs['event_title'],
+                start_time=kwargs['start_time'],
+                end_time=kwargs['end_time'],
+                checked_in=True,
+                cancelled=False)
+            event.save()
+        return EventCheckin(event=event)
 
 
 class CancelEvent(graphene.Mutation):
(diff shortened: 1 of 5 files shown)


api/events/schema.py
```diff
@@ -59,7 +50,7 @@ class Arguments:
     event = graphene.Field(Events)
 
     def mutate(self, info, **kwargs):
-        room_id = Events.check_status(self, info, **kwargs)
+        room_id = RoomSchedules().check_event_status(info, **kwargs)
         cancelled_event = EventsModel(
```
kampkelly: I have a suggestion, I think it will be better if you get the room_id first before calling the `RoomSchedules().check_event_status` method.
I believe it is better that method should return a boolean (true or false).
Reason is `check_event_status` is returning room_id which is quite different from the name of the method.

C3real-kill3r: The calendar_id for the room comes from android when they hit this endpoint. So what I did when I want to populate events model the data doesn't come with the id of the room from the database, (the primary key of the room from the room model). So I used the calendar_id to make a query to the rooms model to get the room_id. And as far as the name of the method goes, I believe is accurate from what it does, it's responsible for giving back the status of the event incase of an error.

C3real-kill3r: And this is canceled event, there's not much Ive done here. test check-in instead @kampkelly

kampkelly: Okay, understood. Please try and improve the test coverage.

C3real-kill3r: @kampkelly done, thanks for the feedback

helpers/calendar/events.py
```diff
@@ -90,3 +92,25 @@ def get_all_room_schedules(self, query, start_date, end_date):
                 all_events.append(current_event)
 
         return all_events, all_dates
+
+    def check_event_status(self, info, **kwargs):
+        try:
+            room_id = RoomModel.query.filter_by(
```
RonKbS: Great work @C3real-kill3r.

Kindly share the reason for using .first().
Rooms are to have unique calender_ids so the data returned here is not likely to be more than one.

fixtures/events/event_checkin_fixtures.py
```diff
@@ -98,3 +98,42 @@
         }
     }
 }
+
+event_existing_in_db_checkin_mutation = '''mutation {
```
RonKbS: I would suggest going with `check_in_mutation_fot_existing_event_in_db`

helpers/calendar/events.py
C3real-kill3r: Yes, rooms have unique ids, but without the .first() I'll get an attribute error since the query wont have the attribute id. So to bypass that I added the function. @RonKbS

fixtures/events/event_checkin_fixtures.py
C3real-kill3r: I thought this was also descriptive enough, but that is also a good one @RonKbS , I'll go with yours.....done @RonKbS

helpers/calendar/events.py
```diff
@@ -90,3 +92,25 @@ def get_all_room_schedules(self, query, start_date, end_date):
...
+                calendar_id=kwargs['calendar_id']).first().id
+            if EventsModel.query.filter_by(
```
Bonifase: Why do we need to use the .count() method yet we are just querying one event object?

C3real-kill3r: without the count, it'll return the query, which will always be true regardless of what it returns. so to get the object I added the count() and checked if it appear more than twice. The other alternative would be to use .one() but that would raise an error of `No row was found` when the query returns no rows. And the failing tests are not on my branch, I'll have to rerun the build till they pass.... @Bonifase what would suggest I use instead?

C3real-kill3r: @Bonifase the tests are fixed

C3real-kill3r:
> I would suggest you reate a variable called checked_in_event, assign it to the query then do the check, same to the canceled event.

noted @Bonifase feedback implemented.

dnuwa:
good work @caveinn. I've tested it out and it meets the PT board story requirements. Though I'd like to propose that the `event_id` be passed as an integer for consistency.

C3real-kill3r:
> good work @caveinn. I've tested it out and it meets the PT board story requirements. Though I'd like to propose that the `event_id` be passed as an integer for consistency.

event_id comes a string @dnuwa, an example of an `event_id` is `4vbcme4nhc0mge0v5id0cic4nj`. So I don't think passing it as an integer will work.

PR #248, opened 2019-02-12
#163849509 Update Firebase Token

michael-basweti:
#### What does this PR do?
* Create an endpoint for updating firebase token. 
*  Added a mutation to `api/rooms/schema.py` to take care of updating the token.
* Added tests for the same.
* git checkout to `ft-update-firebase-token-endpoint-163849509` and run the following query:
## query for updating firebase token
```
 mutation{
  updateFirebaseToken(firebaseToken:"hakjacjihihaehcfkuqehihjfieq-nanana",roomId:1){
    room{
      firebaseToken
      
    }
  }
}

```
## expected response
```
{
  "data": {
    "updateFirebaseToken": {
      "room": {
        "firebaseToken": "hakjacjihihaehcfkuqehihjfieq-nanana"
      }
    }
  }
}
```
* Not applicable
#### screenshots:
* Most booked rooms
screenshot 2019-02-11 at 23 15 40

[#163849509](https://www.pivotaltracker.com/n/projects/2154921/stories/163849509)

### DIFF
--- a/api/room/schema.py
+++ b/api/room/schema.py
@@ -1,7 +1,9 @@
+import requests
 import graphene
 from sqlalchemy import func
 from graphene_sqlalchemy import (SQLAlchemyObjectType)
 from graphql import GraphQLError
+from config import Config
 from api.room.models import Room as RoomModel
 from api.tag.models import Tag as TagModel
 from api.office.models import Office
@@ -183,7 +185,28 @@ class DeleteRoom(graphene.Mutation):
         return DeleteRoom(room=exact_room)
 
 
+class UpdateFirebaseToken(graphene.Mutation):
+
+    class Arguments:
+        room_id = graphene.Int(required=True)
+        firebase_token = graphene.String(required=True)
+    room = graphene.Field(Room)
+
+    def mutate(self, info, room_id, **kwargs):
+        validate_empty_fields(**kwargs)
+        query_room = Room.get_query(info)
+        active_rooms = query_room.filter(RoomModel.state == "active")
+        room = active_rooms.filter(RoomModel.id == room_id).first()
+        if not room:
+            raise GraphQLError("Room not found")
+        update_entity_fields(room, **kwargs)
+        room.save()
+        requests.get(url=Config.MRM_PUSH_URL, params="hello")
+        return UpdateFirebaseToken(room=room)
+
+
 class Mutation(graphene.ObjectType):
     create_room = CreateRoom.Field()
     update_room = UpdateRoom.Field()
     delete_room = DeleteRoom.Field()
+    update_firebase_token = UpdateFirebaseToken.Field()
--- a/config.py
+++ b/config.py
@@ -12,6 +12,8 @@ class Config:
     MAIL_USERNAME = os.getenv('MAIL_USERNAME')
     MAIL_PASSWORD = os.getenv('MAIL_PASSWORD')
     DOMAIN_NAME = os.getenv('DOMAIN_NAME')
+    # mrm_push url
+    MRM_PUSH_URL = os.getenv("MRM_PUSH_URL")
 
     # Celery configuration
     CELERY_BROKER_URL = os.getenv('CELERY_BROKER_URL')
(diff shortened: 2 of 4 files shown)


api/room/schema.py
```diff
@@ -183,7 +184,29 @@ def mutate(self, info, room_id, **kwargs):
...
 
+class UpdateFirebaseToken(graphene.Mutation):
+
+    class Arguments:
+        room_id = graphene.Int(required=True)
+        firebase_token = graphene.String()
+    room = graphene.Field(Room)
+
+    def mutate(self, info, room_id, **kwargs):
+        MRM_URL = "https://www.google.com/"
```
RachaelNantale: Good work @michael-basweti, kindly explain why you're using the google url as the `MRM_URL`

michael-basweti: that's just to show that after updating the token we can hit the mrm_push url and notify the service of the change. once the mrm push is done, we will have the url in the .env and and the above code will be edited so that the service can work. that's just an example

michael-basweti:
> The PT Board story you added doesn't match the PR raised. Please look into this

I had confused it but it has been rectified

michael-basweti:
> This seems to implement exactly what the PT story specifies. I also think you should inform devops to add an env that you can use for the MRM_URL instead of hard-coding a temporary variable.

I have notified them. It was previously in my .env file but my tests failed on circleci, once the field is added, this will be taken care of

api/room/schema.py
```diff
@@ -183,7 +184,29 @@ def mutate(self, info, room_id, **kwargs):
         return DeleteRoom(room=exact_room)
 
 
+class UpdateFirebaseToken(graphene.Mutation):
+
+    class Arguments:
+        room_id = graphene.Int(required=True)
+        firebase_token = graphene.String()
```
RonKbS: Great work @michael-basweti,

Please share your thoughts on adding `required=True` on the firebase_token variable.
I think it is necessary seeing as the mutation requires it to implement the change.

api/room/schema.py
```diff
@@ -183,7 +184,29 @@ def mutate(self, info, room_id, **kwargs):
...
+    room = graphene.Field(Room)
+
+    def mutate(self, info, room_id, **kwargs):
+        MRM_URL = "https://www.google.com/"
+        validate_empty_fields(**kwargs)
+        query_room = Room.get_query(info)
+        active_rooms = query_room.filter(RoomModel.state == "active")
+        room = active_rooms.filter(RoomModel.id == room_id).first()
```
RonKbS: Also, is the .first() required here seeing as the room_ids are unique?

api/room/schema.py
joshuaocero: @michael-basweti kindly read the URL from the `.env` file.

api/room/schema.py
michael-basweti: > Great work @michael-basweti,
> 
> Please share your thoughts on adding `required=True` on the firebase_token variable.
> I think it is necessary seeing as the mutation requires it to implement the change.
I will include it, thanks for that view

api/room/schema.py
michael-basweti: let me test it out

michael-basweti: > Also, is the .first() required here seeing as the room_ids are unique?

@RonKbS , Without the .first() it returns a Query and not an object. We actually need the object in this case.

api/room/schema.py
michael-basweti: @joshuaocero @philophilo is currently implementing that and I will include it once it's done

kampkelly:
Good work @michael-basweti

api/room/schema.py
Bonifase: @michael-basweti Has the URL been updated?

michael-basweti:
@Bonifase @philophilo is still working on it.

api/room/schema.py
```diff
@@ -183,7 +185,29 @@ def mutate(self, info, room_id, **kwargs):
...
 
+class UpdateFirebaseToken(graphene.Mutation):
+
+    class Arguments:
+        room_id = graphene.Int(required=True)
+        firebase_token = graphene.String(required=True)
+    room = graphene.Field(Room)
+
+    def mutate(self, info, room_id, **kwargs):
+        MRM_PUSH_URL = os.getenv("MRM_PUSH_URL")
```
joshuaocero: Please use the config.py file to access environment vars then rebase so we can merge this.

kampkelly:
Tested and it works, please work on @joshuaocero feedback.

PR #250, opened 2019-02-12
#163448618 Implement clean up of deleted records from the db periodically

caveinn:
#### What does this PR do?
- Implement the permanent deletion of data that had been flagged for deletion in the database.

- Implemented the deletion of data that had been flagged for deletion by providing a file that can be run whenever the admin requires a database clean up.

- Pull this branch
- Ensure you have data in the database that has been flagged for deletion (by having state set to deleted) 
- Run the command `python run_mrm_services.py cleandb`
- Check to confirm that the data that had been flagged for deletion has been permanently deleted from the database.

- When permanently deleting, if the data to be deleted has any children they will also be deleted in a cascading manner.

[#163448618 ](https://www.pivotaltracker.com/story/show/163448618)
- N/A
#### Questions:


alembic/versions/ab83c84a325b_include_cascading_deletion.py
```diff
@@ -0,0 +1,86 @@
...
+revision = 'ab83c84a325b'
+down_revision = 'eaca44fbeabe'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_constraint('blocks_office_id_fkey', 'blocks', type_='foreignkey')
+    op.create_foreign_key(None, 'blocks', 'offices', ['office_id'], ['id'], ondelete='CASCADE')
```
michael-basweti: Why did we add `ondelete='CASCADE'` here on the alembic file other than using the one on the model. I think the one on the model should work.

caveinn: The reason for this is that the cascade="all,delete" does not delete the children, it instead removes the relation by setting the foreign key as null. When I tried running the script without the changes the database threw an error that the foreign key constraint was infringed on. This was the solution I found after doing some research.

dnuwa:
good work @caveinn though you need to update your commit. 
`[WIP #163448618]`

caveinn:
> good work @caveinn though you need to update your commit.
> `[WIP #163448618]`

Thanks for the catch. I have updated that.

alembic/versions/ab83c84a325b_include_cascading_deletion.py
RonKbS: @caveinn , Have you tried `all, delete-orphan` in the models as well?

caveinn: @RonKbS I tried that too. After reading the documentation I learned that the cascade attribute on its own does not delete the children it disassociates it. You have to couple them with the ondelete="cascade" to delete the children.

caveinn:
> Is your current implementation able to periodically flush out deleted records from the system and is the admin able to set intervals for when the deletion will automatically happen as required on the story? https://www.pivotaltracker.com/story/show/163448618

Currently the implementation is such that the admin will run the service when they need the data cleaned. The same way the auto cancel was implemented. This was arrived at after some discussion with the TTL who had requested the story.

Bonifase:
> > Is your current implementation able to periodically flush out deleted records from the system and is the admin able to set intervals for when the deletion will automatically happen as required on the story? https://www.pivotaltracker.com/story/show/163448618
> 
> @Bonifase Currently the implementation is such that the admin will run the service when they need the data cleaned. The same way the auto cancel events was implemented. This was arrived at after some discussion with the TTL who had requested the story.

@joshuaocero Kindly confirm this.

caveinn:
> > Is your current implementation able to periodically flush out deleted records from the system and is the admin able to set intervals for when the deletion will automatically happen as required on the story? https://www.pivotaltracker.com/story/show/163448618
> 
> @Bonifase Currently the implementation is such that the admin will run the service when they need the data cleaned. The same way the auto cancel events was implemented. This was arrived at after some discussion with the TTL who had requested the story.

@Bonifase , The original idea was to have a bash script that will run all the services but I hit a snag and followed the above implementation. I contacted the TTL regarding the same and informed him of the implementation I currenlty have. We are supposed to have a sync once he is available. Currently he is putting out some fires in his own words.

joshuaocero:
> > > Is your current implementation able to periodically flush out deleted records from the system and is the admin able to set intervals for when the deletion will automatically happen as required on the story? https://www.pivotaltracker.com/story/show/163448618
> > 
> > 
> > @Bonifase Currently the implementation is such that the admin will run the service when they need the data cleaned. The same way the auto cancel events was implemented. This was arrived at after some discussion with the TTL who had requested the story.
> 
> @joshuaocero Kindly confirm this.

Yes, that is right. The service will be run whenever the data needs to be cleaned. We'll automate it using a cronjob that will be configured by the DevOps team. Great job @caveinn!

run_mrm_services.py
```diff
@@ -0,0 +1,28 @@
+import sys
+from services.room_cancelation.auto_cancel_event import UpdateRecurringEvent
+from services.data_deletion.clean_deleted_data_from_db import DataDeletion
+
+services = {
+    "cleandb": DataDeletion().clean_deleted_data,
```
Bonifase: Avoid using abbreviations for your variable naming

services/data_deletion/clean_deleted_data_from_db.py
```diff
@@ -0,0 +1,23 @@
...
+        This method deletes data that has been marked for
+        deletion from the database
+        """
+        print("cleaning data...")
+        database_engine = create_engine(database_uri)
+        metadata = MetaData()
+        metadata.reflect(bind=database_engine)
+
+        for table in reversed(metadata.sorted_tables):
+            # for each table delete all data marked for deletion
```
Bonifase: KIndly include your comments immediately after the function definition. Use docstrings https://www.python.org/dev/peps/pep-0008/#block-comments

run_mrm_services.py
```diff
@@ -0,0 +1,28 @@
...
+
+def run_mrm_services(arguments):
+    all_services = list(services.keys())
+    try:
+        if not arguments:
+            arguments = list(services.keys())
+        for argument in arguments:
+            services[argument]()
+    except KeyError as wrong_key:
+        print(
```
Bonifase: Why do you prefer print statements in place of return statements?

caveinn: This is because it is a script that will be running on the terminal, hence the best way to inform the user of what is happening is to print it to the terminal.

caveinn:
@Bonifase I have implemented the feedback you left.

PR #253, opened 2019-02-13
#163750909 Fix soft delete bug

RachaelNantale:
#### What does this PR do?
The PR ensures that the user can still create a field with the same name in the case that the older field is archived.

Because we are not deleting rows but simply changing their state, we have scenarios where a user adding a record with say the same name as one that has already been 'archived' fails with a duplication error. The goal to ensure users can add fields even though names are similar as long as only one of them is 'active'.

- Pull the branch `bg-fix-soft-delete-163750909`
- Run migrations `make-migrate`
- run the command `make run-app`
- Run the `createLocation` query
- Run the following query

``` mutation {
  createLocation(name: "Kampala", abbreviation: "KLA", country: "Uganda", timeZone: "EAST_AFRICA_TIME", imageUrl:"https://lala.com") {   
    location {
      name
    }
  }
}
```
```
mutation{
    deleteLocation(locationId:1){
        location{
        id
        }
    }
    }
```

After the field (in this case Kampala ) has been soft-deleted,  use the `createLocation` mutation again and you should be able to create `Kampala` as a new location.

 So much as one can delete and re-create a field with the same name, we cannot have two active fields with the same name. It will give an error.

[#163750909](https://www.pivotaltracker.com/story/show/163750909)

screen shot 2019-02-13 at 2 52 27 pm

#### Then the Delete
screen shot 2019-02-13 at 2 51 48 pm

#### Then the Re-creation
screen shot 2019-02-13 at 2 52 27 pm

#### This  shows the response given when a user tries to create an already active existing field
screen shot 2019-02-13 at 2 52 39 pm

#### This shows that the two fields with the same name but one is active
screen shot 2019-02-13 at 2 58 23 pm


helpers/auth/error_handler.py
```diff
@@ -27,3 +28,33 @@ def __enter__(self):
 
     def __exit__(self, type, value, traceback):
         return False
+
+
+class CheckDuplicationManager():
```
Bonifase: Great work here! Is it possible for you to do these checks for name and state in the SaveContextManager class above? This could save you the time of writing a new entire class to do the same.

RachaelNantale: Thanks, @Bonifase,  it has not been possible so far to have both managers combined.
The `SaveContextManager` checks for an error message from the database. It does not consider the state of the data being sent.

The `CheckDuplicationManager` on the other hand checks for the name and the state without using the error messages from the database.

So the logic used in the two managers is different and combining them has been unsuccessful as we had seen in the sync I had earlier with you.

RachaelNantale:
> Great work here, fix the failing tests. For most-booked and list booked tests, go to `fixtures/room/most_booked_room_least_booked_rooms_fixtures.py` and in each of the responses, change the number of meetings from 29 to 30. Then you can now work on your tests

Thanks @michael-basweti, This has been worked on

fixtures/office/office_fixtures.py
```diff
@@ -194,3 +194,4 @@
     }
 }
 '''
+office_mutation_query_response = {'data': {'createOffice': {'office': {'name': 'The Crest', 'locationId': 1, 'blocks': [{'id': '3', 'name': 'The Crest'}]}}}} # noqa
```
kampkelly: You can add a new line here at the end of the file.

kampkelly:
Good work @RachaelNantale

dnuwa:
Good work @RachaelNantale works as expected.

PR #254, opened 2019-02-14
#163849430 Refactor Cancel Event

nakatuddesuzan:
#### What does this PR do?
- Refactor cancel event
- Make the cancel feature more robust by:
    - Only updating canceled field to True if the event exists in the DB
    - Add the event to the DB if it does not exist with the canceled set to True
- git pull branch `ch-refactor-cancel-event-163849430` and check it out
- use this query to retrieve all events and get you event_id
```
query{
 analyticsForDailyRoomEvents(startDate:"Feb 14 2019", endDate:"Feb 15 2019"){
   day
   events{
    eventId
    eventSummary
     eventSummary
     startTime
     endTime
     roomName
     noOfParticipants
   }
 }
}
```
- run this mutation below to cancel the event
    - remember to update the mutation below with your event details
```
mutation {
        cancelledEvent(
        calendarId:"andela.com_3131313630383336383038@resource.calendar.google.com",
        eventId:"22hojo3qqqr5jdaacr7610qjrg", 
        eventTitle:"suzan<>test",
        startTime:"Feb 15 2019",
        endTime:"Feb 15 2019")
        {
            event{
                eventId
                roomId
                checkedIn
                cancelled
                room{
                    id
                    name
                    calendarId
                }
            }
        }
    }
``` 
- check your database to see if the event is saved
- Before, If an event would be canceled and did not exist in the database, It wouldn't save its self in the database. This has been solved.
[#163849430](https://www.pivotaltracker.com/story/show/163849430)
#### Screenshots
- querying all events
screen shot 2019-02-15 at 12 11 08

- canceling an event
screen shot 2019-02-19 at 12 39 48

- event saved in the database
screen shot 2019-02-15 at 15 18 58

### DIFF
--- a/api/events/schema.py
+++ b/api/events/schema.py
@@ -20,14 +20,8 @@ class EventCheckin(graphene.Mutation):
     event = graphene.Field(Events)
 
     def mutate(self, info, **kwargs):
-        room_id = RoomSchedules().check_event_status(info, **kwargs)
-        event = EventsModel.query.filter_by(
-            start_time=kwargs['start_time'],
-            event_id=kwargs['event_id']).scalar()
-        if event is not None:
-            event.checked_in = True
-            event.save()
-        else:
+        room_id, event = check_event_in_db(self, info, "checked_in", **kwargs)
+        if not event:
             event = EventsModel(
                 event_id=kwargs['event_id'],
                 room_id=room_id,
@@ -50,20 +44,37 @@ class CancelEvent(graphene.Mutation):
     event = graphene.Field(Events)
 
     def mutate(self, info, **kwargs):
-        room_id = RoomSchedules().check_event_status(info, **kwargs)
-        cancelled_event = EventsModel(
-            event_id=kwargs['event_id'],
-            room_id=room_id,
-            event_title=kwargs['event_title'],
-            start_time=kwargs['start_time'],
-            end_time=kwargs['end_time'],
-            checked_in=False,
-            cancelled=True)
-        cancelled_event.save()
+        room_id, event = check_event_in_db(self, info, "cancelled", **kwargs)
+        if not event:
+            event = EventsModel(
+                event_id=kwargs['event_id'],
+                room_id=room_id,
+                event_title=kwargs['event_title'],
+                start_time=kwargs['start_time'],
+                end_time=kwargs['end_time'],
+                checked_in=False,
+                cancelled=True)
+            event.save()
 
-        return CancelEvent(event=cancelled_event)
+        return CancelEvent(event=event)
+
+
+def check_event_in_db(instance, info, event_check, **kwargs):
+    room_id = RoomSchedules().check_event_status(info, **kwargs)
+    event = EventsModel.query.filter_by(
+        start_time=kwargs['start_time'],
+        event_id=kwargs['event_id']).scalar()
+    if event and event_check == 'cancelled':
+        event.cancelled = True
+        event.save()
+        return room_id, event
+    elif event and event_check == 'checked_in':
+        event.checked_in = True
+        event.save()
+        return room_id, event
+    return room_id, event
 
 
 class Mutation(graphene.ObjectType):
     event_checkin = EventCheckin.Field()
-    cancelled_event = CancelEvent.Field()
+    cancel_event = CancelEvent.Field()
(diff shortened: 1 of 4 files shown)


dnuwa:
@nakatuddesuzan kindly update this commit message. 
image

nakatuddesuzan:
Hey @dnuwa , thanks for the concern. Am going to update it after working on all the feedback.

caveinn:
@nakatuddesuzan  The query you have described in the PR  to get all events does  not return eventId.

nakatuddesuzan:
Hey @caveinn , I had implemented the cancel in such a way that it would cancel would cancel the event on the calendar. When I had a chat with @joshuaocero and Taracha about it and was told that the cancel functionality was implemented from the android side. So all I needed to do is to make sure that if an event is not in the DB, then it should be saved with `cancelled=True`.

nakatuddesuzan:
Hello @Bonifase, I haven't changed the name of the mutation, It is still the same(CancelEvent). Kindly look at it again.

nakatuddesuzan:
Hello @Bonifase , If you mean the one in the PR description, I didn't change it too, that is how it is in the fixtures, I just used the same. Maybe if it is something I need to look into. Though I can't find how I can change it since `CancelEvent` throws an error on insomnia. The mutation in the code is still `CancelEvent`. Kindly advise

Bonifase:
> Hello @Bonifase , If you mean the one in the PR description, I didn't change it too, that is how it is in the fixtures, I just used the same. Maybe if it is something I need to look into. Though I can't find how I can change it since `CancelEvent` throws an error on insomnia. The mutation in the code is still `CancelEvent`. Kindly advise

Kindly rename it to `cancelEvent`. I believe this would be part of refactoring

api/events/schema.py
```diff
@@ -50,18 +44,31 @@ class Arguments:
...
+            event.save()
+
+        return CancelEvent(event=event)
+
 
-        return CancelEvent(event=cancelled_event)
+def check_event_in_db(instance, info, event_check, **kwargs):
+    room_id = RoomSchedules().check_event_status(info, **kwargs)
+    event = EventsModel.query.filter_by(
+        start_time=kwargs['start_time'],
```
joshuaocero: @nakatuddesuzan why must we filter by `start_time`?

Bonifase:
> @nakatuddesuzan kindly update this commit message.
> image

Please work on your commit message

api/events/schema.py
nakatuddesuzan: Hello @joshuaocero , thanks for pointing this out. I don't think we need to filter by start_time. event_id alone can work since its unique.

C3real-kill3r:
> Hi @nakatuddesuzan, I have tested this functionality. However on cancelling the event, it creates a duplicate event, only with the cancelled field as false. The original event still exists. You're supposed to update the original event, not create another. Kindly look into this.

+1 
image

You could have just used the function I created earlier. It will make your work.

nakatuddesuzan:
> > Hi @nakatuddesuzan, I have tested this functionality. However on cancelling the event, it creates a duplicate event, only with the cancelled field as false. The original event still exists. You're supposed to update the original event, not create another. Kindly look into this.
> 
> +1
> image
> 
> You could have just used the function I created earlier. It will make your work.

nakatuddesuzan:
> > Hi @nakatuddesuzan, I have tested this functionality. However on cancelling the event, it creates a duplicate event, only with the cancelled field as false. The original event still exists. You're supposed to update the original event, not create another. Kindly look into this.
> 
> +1
> image
> 
> You could have just used the function I created earlier. It will make your work.

Hello @C3real-kill3r , I used the same function. that other function was just a refactor. I guess I missed out something. Let me look through it

nakatuddesuzan:
Hello @bevkololi and @C3real-kill3r , I have worked on your feedback, The refactor I had made for those two methods wasn't successful that is why I had that bug, but it's now fixed.

api/events/schema.py
nakatuddesuzan: Hello @joshuaocero , on second thought, we need to query by start_time to cater for recurring events. If we don't include start_time we may end up canceling all events with the same `event_id`

PR #256, opened 2019-02-14
#163936513 Add title field to Question Model

dnuwa:
#### What does this PR do?
- Add title field to Question Model and GraphQL endpoints 
- Adds `question_tile` field to the Questions model
- Updates the `CreateQuestion` and `UpdateQuestion` mutation classes
- Updates the questions fixtures
- Generate migration file for the new changes
* git checkout to `ch-add-title-field-to-Question-Model-163936513`
* run migrations
* test it out by running the mutations below
createQuestion
````
mutation{
  createQuestion(questionType:"Rate",
  questionTitle:"Rating",
  question:"How will you rate the brightness of the room",
  startDate:"2019-02-15T13:00:00", endDate:"2019-02-16T13:00:00") {
    question{
      id
      questionTitle
      question
      questionType
      startDate
      endDate
      }
  }
}
````
updateQuestion
````
mutation {
      updateQuestion(questionId:2,
      questionTitle:"Rating Feedback"
      startDate:"2019-02-15T13:00:00", endDate:"2019-02-16T13:00:00") {
        question {
          id,
          questionTitle,
          startDate,
          endDate,
        }
      }
  }
````
N/A
[#163936513](https://www.pivotaltracker.com/story/show/163936513)
#### Screenshots
Create Question 
image

Update Question.
image

Querry all Questions.
image

### CODE REVIEW

File: api/question/schema.py [hunk 1]
```diff
@@ -60,6 +61,7 @@ def resolve_questions(self, info):
 class UpdateQuestion(graphene.Mutation):
     class Arguments:
         question_id = graphene.Int(required=True)
+        question_title = graphene.String(required=True)
```
caveinn: I do not think the question title should be required in the update mutation since it seems you will be forcing the user to update the question_title every time they need to update something.

dnuwa: @caveinn I assume when the question changes, one might want to change the title too. Also since the `question_title` field has been assigned a default value for the questions that are already existing, this is to give the front end team a chance to be able to edit the titles to what they see fit for each question.

dnuwa:
> This looks good and works according to expectations, however, please limit the question to 15 characters the way @Bonifase suggested
@michael-basweti thanx for pointing that out. I take it that you meant the `question_title`. on the screenshot, the text that is circled is an example of a `question_title` and it is about 18 characters. not limiting the number of characters used enables one to create a more meaningful title as the one highlighted. @Bonifase kindly share more on this if you still think the characters need to be limited to 15.
image

dnuwa:
> As @caveinn said, I think setting required to True in the question title is not okay, user may just want to change one field e.g question_type, forcing the user to change the title to might be inconveniencing.

Thanx @caveinn and @bevkololi your concern are very valid and I've implemented it.

api/question/schema.py
dnuwa: @caveinn I had miss understood your concern but it's now clear and it has been rectified.

PR #260, opened 2019-02-15
#163783849 Add cascading effect on soft-delete feature

RonKbS:
#### What does this PR do?
The PR enables the deletion of related data if a delete mutation is used

When deletions are done, related data is not included in the deletion/archiving. As an example, deleting a location will leave the rooms for that particular location still available. This PR enables this to be carried out automatically.

Pull the branch and run a deletion query such as the one of locations below.

``` 
mutation {
  deleteLocation(locationId: 1) {
    location {
      name
    }
  }
}
```

Any offices and rooms related to the location will also be deleted/archived.

N/A

[#163977386](https://www.pivotaltracker.com/story/show/163977386)

N/A

### DIFF
--- a/api/room/models.py
+++ b/api/room/models.py
@@ -6,7 +6,7 @@ from graphql import GraphQLError
 from sqlalchemy.schema import Sequence
 
 from helpers.database import Base, db_session
-from utilities.utility import Utility, StateType
+from utilities.utility import Utility, StateType, cascade_soft_delete
 from api.floor.models import Floor  # noqa: F401
 from api.wing.models import Wing  # noqa: F401
 from api.events.models import Events  # noqa: F401
@@ -73,3 +73,11 @@ def receive_before_insert(mapper, connection, target):
         if not verify_calendar_id(target.calendar_id):
             raise GraphQLError("Room calendar Id is invalid")
         pass
+
+
+cascade_soft_delete(
+    Room, 'room_resource', 'room_id'
+)
+cascade_soft_delete(
+    Room, 'events', 'room_id'
+)
--- a/utilities/utility.py
+++ b/utilities/utility.py
@@ -1,5 +1,7 @@
-from helpers.database import db_session
 import enum
+import api
+from helpers.database import db_session
+from sqlalchemy import event
 
 
 def update_entity_fields(entity, **kwargs):
@@ -12,6 +14,41 @@ def update_entity_fields(entity, **kwargs):
     for key in keys:
         exec("entity.{0} = kwargs['{0}']".format(key))
     return entity
+
+
+def cascade_soft_delete(parent_model, child, parent_id):
+    @event.listens_for(parent_model, 'after_update')
+    def receive_after_update(mapper, connection, target):
+        listen_for_after_flush_event(target, child, parent_id)
+
+
+def listen_for_after_flush_event(target, child, parent_id):
+    @event.listens_for(db_session, "after_flush", once=True)
+    def receive_after_flush(session, context):
+        """
+        Function to update related child data after a delete
+        mutation has been run.
+        The check statement on line 42 caters for the difference
+        in naming of room_resource folder and Resource class
+            :param session:
+            :param context:
+        """
+        if target.state == 'archived':
+            child_model = getattr(api, child)
+            child_object = getattr(
+                child_model.models,
+                'Resource' if child.capitalize() == 'Room_resource'
+                else child.capitalize()
+            )
+            id_of_parent = getattr(child_object, parent_id)
+            session.query(
+                child_object
+            ).filter(
+                id_of_parent == target.id
+            ).update(
+                {child_object.state: target.state}
+            )
+        pass
 
 
 class Utility(object):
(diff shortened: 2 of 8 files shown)


utilities/utility.py
```diff
@@ -14,6 +16,30 @@ def update_entity_fields(entity, **kwargs):
...
+    def receive_after_update(mapper, connection, target):
+        @event.listens_for(db_session, "after_flush", once=True)
+        def receive_after_flush(session, context):
+            if target.state == 'archived':
+                child_model = getattr(api, child)
+                # Check for resource name is due to the improper
+                # naming of folder room_resource and class Resource
+                child_object = getattr(
+                    child_model.models,
+                    'Resource' if child.capitalize() == 'Room_resource' else child.capitalize()  # noqa: E501
```
caveinn: I think you  can remove then noqa by having the if on statement on different lines or having a child.capitilize() assigned to a variable with a shorter name.

utilities/utility.py
```diff
@@ -14,6 +16,30 @@ def update_entity_fields(entity, **kwargs):
...
+                )
+                parent_s_id = getattr(child_object, parent_id)
```
caveinn: I think parents_id would be more intuitive the s seems misplaced in the above name.

api/room/models.py
```diff
@@ -59,3 +59,11 @@ def receive_after_flush(session, context):
...
             raise GraphQLError("Room calendar Id is invalid")
         pass
+
+
+cascade_soft_delete(
+    Room, 'room_resource', 'room_id'
+)
+cascade_soft_delete(
+    Room, 'events', 'room_id'
+)
```
bevkololi: Kindly include devices and responses in soft-deletion of room responses.

utilities/utility.py
RonKbS: parents_id would refer to more than one parent. 
parent_s_id replaces ' in parent's hence its use here.

api/room/models.py
RonKbS: Responses and devices have no state column in their models so calling this function on them would not work.

As discussed in the backend-sync, stories are to be raised to fix this.

utilities/utility.py
```diff
@@ -14,6 +16,30 @@ def update_entity_fields(entity, **kwargs):
     return entity
 
 
+def cascade_soft_delete(parent_model, child, parent_id):
+    @event.listens_for(parent_model, 'after_update')
```
michael-basweti: @RonKbS is there a way we can have the two wrappers in one overall function, this way we can reduce complexity by reducing the number of functions.

RonKbS: I have just pushed changes with one each decorator in its own function and this should solve the complexity issue @michael-basweti.

Thanks for the feedback.

utilities/utility.py
```diff
@@ -14,6 +16,35 @@ def update_entity_fields(entity, **kwargs):
...
+    def receive_after_update(mapper, connection, target):
+        listen_for_after_flush_event(target, child, parent_id)
+
+
+def listen_for_after_flush_event(target, child, parent_id):
+    @event.listens_for(db_session, "after_flush", once=True)
+    def receive_after_flush(session, context):
+        if target.state == 'archived':
+            child_model = getattr(api, child)
+            # Check for resource name is due to the improper
```
Bonifase: KIndly include your comments immediately after the function definition. Use docstrings https://www.python.org/dev/peps/pep-0008/#block-comments

utilities/utility.py
```diff
@@ -14,6 +16,35 @@ def update_entity_fields(entity, **kwargs):
...
+            # naming of folder room_resource and class Resource
+            child_object = getattr(
+                child_model.models,
+                'Resource' if child.capitalize() == 'Room_resource'
+                else child.capitalize()
+            )
+            parent_s_id = getattr(child_object, parent_id)
```
Bonifase: Did you mean `parents_id`? If not so then use a more descriptive variable naming

RonKbS:
The recommended changes have been implemented @Bonifase.

utilities/utility.py
```diff
@@ -14,6 +16,41 @@ def update_entity_fields(entity, **kwargs):
...
+        The check statement on line 42 caters for the difference
+        in naming of room_resource folder and Resource class
+            :param session:
+            :param context:
+        """
+        if target.state == 'archived':
+            child_model = getattr(api, child)
+            child_object = getattr(
+                child_model.models,
+                'Resource' if child.capitalize() == 'Room_resource'
```
nakatuddesuzan: Hello @RonKbS , good work. help me understand more about this code by explaining why there is a need to use the capitalize method here. Thanks

RonKbS: @nakatuddesuzan, the method needs to run queries on an object like `api.room.models.Room` and to make it generic one string is passed in (`room`) that is then capitalized to `Room`

RonKbS:
> Great work on the cascade I've tested this and it works @RonKbS, Kindly fix the merge conflicts.

Conflicts have been resolved @RachaelNantale.

dnuwa:
tested and works fine. good work @RonKbS

PR #264, opened 2019-02-19
#163754403 Display friendly message when app crashes due to database errors

C3real-kill3r:
#### What does this PR do?
- [x] catch all sqlalchemy ProgrammingErrors
- [x] catch InternalErrors when server is down
- [x] display user friendly message for respective errors

Display user-friendly messages such as "The database cannot be reached" when the app crashes due to a database error. 

- Pull this branch
- make the migrations as explained in the [README](https://github.com/andela/mrm_api/blob/develop/Readme.md)
- stop the server and try to run any mutation/query(to test for InternalError)
- delete a relation from the database and run a mutation that depends on that relation( to test for ProgrammingError)

### Below are the screenshots 
* initial error message for InternalErrors
screenshot 2019-02-19 at 16 37 02

* new message for InternalErrors
screenshot 2019-02-19 at 16 55 45

* initial message for programmingErrors
screenshot 2019-02-19 at 11 31 17

* new error message for programmingErrors
screenshot 2019-02-22 at 15 13 24

`InternalError` is sometimes raised by drivers in the context of the database connection being dropped, or not being able to connect to the database.
ProgrammingError - Exception raised for programming errors, e.g. table not found or already exists, syntax error in the SQL statement, wrong number of parameters specified, etc.

[#163754403](https://www.pivotaltracker.com/story/show/163754403)

### DIFF
--- a/api/block/schema.py
+++ b/api/block/schema.py
@@ -1,7 +1,7 @@
 import graphene
 from graphql import GraphQLError
 from graphene_sqlalchemy import SQLAlchemyObjectType
-from sqlalchemy import func
+from sqlalchemy import func, exc
 from api.block.models import Block as BlockModel
 from api.room.schema import Room
 from api.office.models import Office
@@ -9,6 +9,7 @@ from api.room.models import Room as RoomModel
 from helpers.room_filter.room_filter import room_join_location
 from helpers.auth.admin_roles import admin_roles
 from utilities.validations import validate_empty_fields
+from utilities.validator import ErrorHandler
 from utilities.utility import update_entity_fields
 from helpers.auth.authentication import Auth
 from helpers.auth.error_handler import SaveContextManager
@@ -28,22 +29,34 @@ class CreateBlock(graphene.Mutation):
 
     @Auth.user_roles('Admin')
     def mutate(self, info, **kwargs):
-        validate_empty_fields(**kwargs)
-        get_office = Office.query.filter_by(id=kwargs['office_id']).first()
-        if not get_office:
-            raise GraphQLError("Office not found")
-        location = get_office.location.name
-        if location.lower() == 'nairobi':
-            block = BlockModel(**kwargs)
-            payload = {
-                'model': BlockModel, 'field': 'name', 'value':  kwargs['name']
-            }
-            with SaveContextManager(
-                block, 'Block', payload
-            ):
-                return CreateBlock(block=block)
-        else:
-            raise GraphQLError("You can only create block in Nairobi")
+        try:
+            validate_empty_fields(**kwargs)
+            get_office = Office.query.filter_by(id=kwargs['office_id']).first()
+            if not get_office:
+                raise GraphQLError("Office not found")
+            location = get_office.location.name
+            if location.lower() == 'nairobi':
+                block = BlockModel(**kwargs)
+                payload = {
+                    'model': BlockModel, 'field': 'name',
+                    'value':  kwargs['name']
+                }
+                query = Block.get_query(info)
+                result = query.filter(BlockModel.state == "active")
+                block_name = result.filter(
+                    func.lower(BlockModel.name) ==
+                    func.lower(kwargs['name'])).count()
+                if block_name > 0:
+                    ErrorHandler.check_conflict(self, kwargs['name'], 'Block')
+                with SaveContextManager(
+                    block, 'Block', payload
+                ):
+                    return CreateBlock(block=block)
+            else:
+                raise GraphQLError("You can only create block in Nairobi")
+        except exc.ProgrammingError:
+            raise GraphQLError("There seems to be a database connection error, \
+                contact your administrator for assistance")
 
 
 class UpdateBlock(graphene.Mutation):
@@ -55,18 +68,22 @@ class UpdateBlock(graphene.Mutation):
 
     @Auth.user_roles("Admin")
     def mutate(self, info, block_id, **kwargs):
-        validate_empty_fields(**kwargs)
-        query = Block.get_query(info)
-        result = query.filter(BlockModel.state == "active")
-        exact_block = result.filter(BlockModel.id == block_id).first()
-        if not exact_block:
-            raise GraphQLError("Block not found")
+        try:
+            validate_empty_fields(**kwargs)
+            query = Block.get_query(info)
+            result = query.filter(BlockModel.state == "active")
+            exact_block = result.filter(BlockModel.id == block_id).first()
+            if not exact_block:
+                raise GraphQLError("Block not found")
 
-        admin_roles.create_floor_update_delete_block(block_id)
+            admin_roles.create_floor_update_delete_block(block_id)
 
-        update_entity_fields(exact_block, **kwargs)
-        exact_block.save()
-        return UpdateBlock(block=exact_block)
+            update_entity_fields(exact_block, **kwargs)
+            exact_block.save()
+            return UpdateBlock(block=exact_block)
+        except exc.ProgrammingError:
+            raise GraphQLError("There seems to be a database connection error, \
+                contact your administrator for assistance")
 
 
 class DeleteBlock(graphene.Mutation):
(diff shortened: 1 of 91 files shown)


api/block/schema.py
```diff
@@ -27,20 +27,23 @@ class Arguments:
...
-        location = get_office.location.name
-        if location.lower() == 'nairobi':
-            block = BlockModel(**kwargs)
-            block.save()
-        else:
-            raise GraphQLError("You can only create block in Nairobi")
-        return CreateBlock(block=block)
+        try:
+            validate_empty_fields(**kwargs)
+            block_name = BlockModel.query.filter_by(name=kwargs['name'].title()).all()  # noqa
```
caveinn: Try avoiding noqa here.

api/role/schema.py
```diff
@@ -18,9 +21,17 @@ class Arguments:
...
     def mutate(self, info, **kwargs):
-        role = RoleModel(**kwargs)
-        with SaveContextManager(role, kwargs.get('role'), 'Role'):
-            return CreateRole(role=role)
+        try:
+            role = RoleModel(**kwargs)
+            query = Role.get_query(info)
+            role_name = query.filter(
+                func.lower(RoleModel.role) == func.lower(kwargs.get('role')))
+            if role_name.count() > 0:
```
caveinn: I think for this you can just have if role_name.count() and leave out the > 0 part.

dnuwa:
@C3real-kill3r kindly work on the conflicts.

api/role/schema.py
C3real-kill3r: Done @caveinn

api/block/schema.py
C3real-kill3r: Done @caveinn

api/role/schema.py
```diff
@@ -18,25 +21,43 @@ class Arguments:
...
-            }
-        with SaveContextManager(role,  'Role', payload):
-            return CreateRole(role=role)
+        try:
+            role = RoleModel(**kwargs)
+            payload = {
+                'model': RoleModel, 'field': 'role', 'value':  kwargs['role']
+                }
+            query = Role.get_query(info)
+            role_name = query.filter(
```
RachaelNantale: kindly remove this code because its redundant with the presence of the context manager

api/tag/schema.py
```diff
@@ -22,15 +24,24 @@ class Arguments:
...
-            return CreateTag(tag=tag)
+        try:
+            validate_empty_fields(**kwargs)
+            tag = TagModel(**kwargs)
+            payload = {
+                'model': TagModel, 'field': 'name', 'value':  kwargs['name']
+                }
+            query = Tag.get_query(info)
+            result = query.filter(TagModel.state == "active")
+            tag_name = result.filter(
```
RachaelNantale: kindly remove this code because its redundant with the presence of the context manager

api/user/schema.py
```diff
@@ -32,36 +33,52 @@ class Arguments:
...
+        try:
+            user = UserModel(**kwargs)
+            if not verify_email(user.email):
+                raise GraphQLError("This email is not allowed")
+            payload = {
+                'model': UserModel, 'field': 'email', 'value':  kwargs['email']
+            }
+            query = User.get_query(info)
+            result = query.filter(UserModel.state == "active")
+            user_name = result.filter(
```
RachaelNantale: kindly remove this code because its redundant with the presence of the context manager

api/wing/schema.py
```diff
@@ -26,20 +27,29 @@ class Arguments:
...
+            if not floor:
+                raise GraphQLError("Floor not found")
+            admin_roles.check_office_location_create_wing(floor_id=kwargs['floor_id'])  # noqa: E501
+            admin_roles.create_update_delete_wing()
+            wing = WingModel(**kwargs)
+            payload = {
+                'model': WingModel, 'field': 'name', 'value':  kwargs['name']
+            }
+            query = Wing.get_query(info)
+            result = query.filter(WingModel.state == "active")
```
RachaelNantale: kindly remove this code because its redundant with the presence of the context manager

api/wing/schema.py
```diff
@@ -72,21 +86,30 @@ class Arguments:
...
+            exact_wing = result.filter(WingModel.id == wing_id).first()
+            if not exact_wing:
+                raise GraphQLError("Wing not found")
+            admin_roles.create_update_delete_wing()
+            update_entity_fields(exact_wing, **kwargs)
+            payload = {
+                'model': WingModel, 'field': 'name', 'value':  kwargs['name']
+            }
+            query = Wing.get_query(info)
+            result = query.filter(WingModel.state == "active")
```
RachaelNantale: kindly remove this code because its redundant with the presence of the context manager

fixtures/questions/get_question_fixtures.py
```diff
@@ -182,3 +182,40 @@
   }
 }
 '''
+
+# response_for_wrong_migration = {
```
RachaelNantale: Please remove the commented out code if its unecessary

api/block/schema.py
```diff
@@ -28,22 +29,34 @@ class Arguments:
...
+                if block_name > 0:
+                    ErrorHandler.check_conflict(self, kwargs['name'], 'Block')
+                with SaveContextManager(
+                    block, 'Block', payload
+                ):
+                    return CreateBlock(block=block)
+            else:
+                raise GraphQLError("You can only create block in Nairobi")
+        except exc.ProgrammingError:
+            raise GraphQLError("There seems to be a database connection error, \
```
Bonifase: Is there a way you could avoid the back slash?

C3real-kill3r: By making the message shorter, that would be the only way @Bonifase

api/role/schema.py
C3real-kill3r: After the implementation of my feature, even after running the migration afresh, the query wasn't giving the intended error message even though it wasn't creating the roles with the same name. So the context manager was preventing creation of roles with duplicate names but it wasn't giving the error message(instead It was showing that the role was created, but in real sense it wasn't), so to give the error message I had to to do the check with this code explicitly. I hope that makes sense @RachaelNantale

api/tag/schema.py
C3real-kill3r: Same as above @RachaelNantale

C3real-kill3r:
> Because a number of developers raised concern that the message would somehow give them hard time in debugging their code when testing, would it be possible to raise the friendly error message only in Production/Staging environment but retain the debugging error in the testing environment?

It's possible but I do not think this is a good idea since it will add to the complexity of the code and will add a lot of code that cannot be tested. If we have a check for code that will only run during production there is no way to run it in testing hence it will forever be uncovered. That's why in the initial implementation I was giving what the error exact

Bonifase:
> > Because a number of developers raised concern that the message would somehow give them hard time in debugging their code when testing, would it be possible to raise the friendly error message only in Production/Staging environment but retain the debugging error in the testing environment?
> 
> It's possible but I do not think this is a good idea since it will add to the complexity of the code and will add a lot of code that cannot be tested. If we have a check for code that will only run during production there is no way to run it in testing hence it will forever be uncovered. That's why in the initial implementation I was giving what the error exact

I don't think this would bring complexity to the code, it's just a matter of checking if the running environment is Production/Staging then raise the custom message. If not, you leave it the way it was before for debugging purposes. Please take your time to think about this.

C3real-kill3r:
> > > Because a number of developers raised concern that the message would somehow give them hard time in debugging their code when testing, would it be possible to raise the friendly error message only in Production/Staging environment but retain the debugging error in the testing environment?
> > 
> > 
> > It's possible but I do not think this is a good idea since it will add to the complexity of the code and will add a lot of code that cannot be tested. If we have a check for code that will only run during production there is no way to run it in testing hence it will forever be uncovered. That's why in the initial implementation I was giving what the error exact
> 
> I don't think this would bring complexity to the code, it's just a matter of checking if the running environment is Production/Staging then raise the custom message. If not, you leave it the way it was before for debugging purposes. Please take your time to think about this.

Okay, lemme test it and ping you

tests/base.py
```diff
@@ -43,6 +43,7 @@ def setUp(self):
         app = self.create_app()
         self.app_test = app.test_client()
         with app.app_context():
+            print(Base.metadata.create_all(bind=engine))
```
nakatuddesuzan: Hey @C3real-kill3r , please remember to remove this print statement If It's not required.

C3real-kill3r: Thanks @nakatuddesuzan noted

api/block/schema.py
nakatuddesuzan: Hey @C3real-kill3r , doe this ```
                raise GraphQLError(
                "There seems to be a database connection error" 
                ", contact your administrator for assistance")
               ``` 
solve the raised issue about the backslash?

C3real-kill3r:
@dnuwa @RachaelNantale @michael-basweti @kampkelly @bevkololi This PR is no longer active, instead refer to this https://github.com/andela/mrm_api/pull/272

PR #270, opened 2019-02-26
#164163073 Test that generated alembic files have no errors.

RonKbS:
#### What does this PR do?
The PR enables the use of newly generated alembic files when testing changes to the codebase.

Previously, any changes to alembic files would go unnoticed during testing because tables were generated using sqlalchemy's `.create_all)` method.
This PR includes commands to run migrations during the setup for each test.
The PR also includes a name generator for constraints in tables such as unique and foreign key.

Check the test db for an alembic_version table after running tests.
Before this implementation, there is no such table created.

N/A

[#164163073](https://www.pivotaltracker.com/story/show/164163073)

N/A

### DIFF
--- a/alembic/env.py
+++ b/alembic/env.py
@@ -21,7 +21,11 @@ import os
 import sys
 
 sys.path.append(os.getcwd())
-config.set_main_option('sqlalchemy.url', os.getenv('DEV_DATABASE_URL'))
+config.set_main_option(
+    'sqlalchemy.url',
+    os.getenv('TEST_DATABASE_URL') if os.getenv('APP_SETTINGS') == 'testing'
+    else os.getenv('DEV_DATABASE_URL')
+)
 
 from helpers.database import Base
 from api.location.models import Location
--- a/.coveragerc
+++ b/.coveragerc
@@ -9,6 +9,7 @@ omit =
 show_missing = True
 omit =
     */python?.?/*
+    */alembic/*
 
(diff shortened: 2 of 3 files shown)


alembic/env.py
```diff
@@ -21,7 +21,11 @@
 import sys
 
 sys.path.append(os.getcwd())
-config.set_main_option('sqlalchemy.url', os.getenv('DEV_DATABASE_URL'))
+config.set_main_option(
+    'sqlalchemy.url',
+    os.getenv('TEST_DATABASE_URL') if os.getenv('APP_SETTINGS') == 'testing'
```
michael-basweti: Nice work @RonKbS I tested and introduced multiple heads intentionally and it worked. Looking at this setting, it caters for only settings and development, what will happen when the code is pushed to production. The environmental variable name for the Production database is `DATABASE_URL`. Can we have a look into this please.

RonKbS: Thank you @michael-basweti. I consulted with @philophilo on this and he said that migrations are run differently on production and so there is no need to include a check for production settings here. As such the check here is for APP_SETTINGS=testing, otherwise it behaves as before.

dnuwa:
awesome work @RonKbS though an example of how multiple heads can be introduced to test its effectiveness would be very helpful.

joshuaocero:
Great job @RonKbS!

PR #280, opened 2019-03-01
#164163073 (V2) Test that generated alembic files have no errors.

RonKbS:
#### What does this PR do?
The PR enables the use of newly generated alembic files when testing changes to the codebase.

Previously, any changes to alembic files would go unnoticed during testing because tables were generated using sqlalchemy's `.create_all)` method.
This PR includes commands to run migrations during the setup for each test.
The PR also includes a name generator for constraints in tables such as unique and foreign key.

Check the test db for an alembic_version table after running tests.
Before this implementation, there is no such table created.

N/A

[#164163073](https://www.pivotaltracker.com/story/show/164163073)

N/A

### DIFF
--- a/alembic/env.py
+++ b/alembic/env.py
@@ -20,8 +20,17 @@ fileConfig(config.config_file_name)
 import os
 import sys
 
+
+database_url = ''
+if os.getenv('APP_SETTINGS') == 'testing':
+    database_url = os.getenv('TEST_DATABASE_URL')
+elif os.getenv('APP_SETTINGS') == 'development':
+    database_url = os.getenv('DEV_DATABASE_URL')
+elif os.getenv('APP_SETTINGS') == 'production':
+    database_url = os.getenv('DATABASE_URL')
+
 sys.path.append(os.getcwd())
-config.set_main_option('sqlalchemy.url', os.getenv('DEV_DATABASE_URL'))
+config.set_main_option('sqlalchemy.url', database_url)
 
 from helpers.database import Base
 from api.location.models import Location
--- a/.coveragerc
+++ b/.coveragerc
@@ -9,6 +9,7 @@ omit =
 show_missing = True
 omit =
     */python?.?/*
+    */alembic/*
 
 [html]
 directory=html_coverage_report
--- a/tests/base.py
+++ b/tests/base.py
@@ -5,6 +5,7 @@ import json
 from flask_testing import TestCase
 from graphene.test import Client
 from datetime import datetime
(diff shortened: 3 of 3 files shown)


joshuaocero:
@RonKbS can you merge the latest changes on v2 into this branch? Let us know how it goes because it looks like everyone that branched off from develop may have to do the same thing.

alembic/env.py
```diff
@@ -21,7 +21,11 @@
 import sys
 
 sys.path.append(os.getcwd())
-config.set_main_option('sqlalchemy.url', os.getenv('DEV_DATABASE_URL'))
+config.set_main_option(
+    'sqlalchemy.url',
+    os.getenv('TEST_DATABASE_URL') if os.getenv('APP_SETTINGS') == 'testing'
+    else os.getenv('DEV_DATABASE_URL')
```
caveinn: I think this will make us use Dev_database even on production. We should also check for the production settings to get the production database.

caveinn: I also think this is better setup on the config file.

RonKbS:
> @RonKbS can you merge the latest changes on v2 into this branch? Let us know how it goes because it looks like everyone that branched off from develop may have to do the same thing.

This has been implemented @joshuaocero

alembic/env.py
RonKbS: > I think this will make us use Dev_database even on production. We should also check for the production settings to get the production database.

I too had similar thoughts regarding this but @philophilo mentioned that migrations are run differently on production. I plan on asking for further clarity however given the error posted on slack over the weekend

RonKbS: > I also think this is better setup on the config file.

The config file would not suit this seeing as the configuration being done here is in regards to alembic which,  if one goes through the documents (https://alembic.sqlalchemy.org/en/latest/api/config.html), requires this to be put in the env.py

kampkelly: > > I think this will make us use Dev_database even on production. We should also check for the production settings to get the production database.
> 
> I too had similar thoughts regarding this but @philophilo mentioned that migrations are run differently on production. I plan on asking for further clarity however given the error posted on slack over the weekend

@RonKbS  any update from @philophilo regarding this?

RonKbS: We are to have a sync later today as he was unavailable in the morning.

kampkelly: Okay.

RonKbS: Following the sync with @philophilo, I have updated this to include a check for production settings as well.

PR #284, opened 2019-03-04
#163971800 (V2) Mock google calendar api for tests

bevkololi:
#### What does this PR do?
* Mock the Google Calendar API calls for the tests.
*  Backend tests currently make direct calls to the Google Calendar API. Mock these API calls so we only hit the live API on Staging and Production.
* The codebase has been refactored to ensure that the api calls are made in one file `helpers/calendar/credentials.py`
* The tests have been refactored to ensure that the data they use is from the files `events.json` and `calendar_list.json`, and no longer hit the live api.
* git checkout to `ft-mock-calendar-api-163971800 ` and run the tests.
* Not applicable
#### screenshots:
* Not applicable

[#163971800](https://www.pivotaltracker.com/story/show/163971800)

### DIFF
--- a/helpers/calendar/credentials.py
+++ b/helpers/calendar/credentials.py
@@ -34,3 +34,20 @@ class Credentials():
         service = build('calendar', 'v3', developerKey=api_key,
                         http=credentials.authorize(Http()))
         return service
+
+
+def get_google_calendar_events(calendarId=None, timeMin=None,
+                               timeMax=None, singleEvents=None, orderBy=None):
+    credentials = Credentials()
+    service = credentials.set_api_credentials()
+    events = service.events().list(calendarId=calendarId, timeMin=timeMin,
+                                   timeMax=timeMax, singleEvents=singleEvents,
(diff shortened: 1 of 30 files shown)


bevkololi:
I had previously used the same branch as in V1 to create this PR. As it has not been merged, I've created the V2 PR using a different branch.

helpers/calendar/credentials.py
```diff
@@ -34,3 +34,28 @@ def set_api_credentials(self):
         service = build('calendar', 'v3', developerKey=api_key,
                         http=credentials.authorize(Http()))
         return service
+
+
+def get_events_within_datetime_range(calendarId=None, timeMin=None,
```
Bonifase: Since the difference between this function and the function defined in line 50 is the arguments passed to them, would it be appropriate to have only one function defined for both of them?

bevkololi: Thanks for pointing this repetition out. It has been worked on.

PR #289, opened 2019-03-04
#164266160 Add node(level) to structure tree

nakatuddesuzan:
#### What does this PR do?
- Add node(level) to structure tree
- implement a method on structure model to add model
- create Graphql endpoint to add a node

- git pull and checkout branch `ft-add-node-to-structure-tree-164266160`
- run this graphql mutation to create the root node
```
mutation{
  createNode(name:"floor"){
    node{
      name
      level
      left
      right
    }
    
  }
}
```
- check the database to check if the root has been created
- note the id of the root
- run the above mutation but this time provide the parent_id which is the root id as shown below
 ```
mutation{
  createNode(name:"floor", parentId:3){
    node{
      name
      level
      left
      right
    }
    
  }
}
```
- the parent_id of the root node is `null`
[#164266160](https://www.pivotaltracker.com/story/show/164266160)
#### Screenshots
screen shot 2019-03-05 at 17 20 53

- create node mutation and output
screen shot 2019-03-05 at 17 09 26

### CODE REVIEW

File: api/office_structure/models.py [hunk 1]
```diff
@@ -17,4 +19,18 @@ class OfficeStructure(Base, Utility, BaseNestedSets):
...
+    def add_node(self, name, parent_id=None):
+        """Function for adding nodes"""
+        if parent_id is None:
+            db_session.add(OfficeStructure(name=name))
+        else:
+            db_session.add(OfficeStructure(name=name, parent_id=parent_id))
+
+    def add_branch(self, nodes):
+        """
+        Function to addbone single branch
```
dnuwa: @nakatuddesuzan kindly fix the typo in this Docstring

nakatuddesuzan:
> @nakatuddesuzan did you fix the parent id issue?

Yes @Bonifase , I did

api/office_structure/schema.py
```diff
@@ -0,0 +1,49 @@
...
+class CreateNode(graphene.Mutation):
+    """ Returns payload on node creation"""
+
+    class Arguments:
+        name = graphene.String(required=True)
+        parent_id = graphene.Int(required=False)
+    node = graphene.Field(OfficeStructure)
+
+    @Auth.user_roles('Admin')
+    def mutate(self, info, **kwargs):
```
michael-basweti: Great work but currently I can create a node with an empty space, maybe you can look at the validate_empty_fields() and see how you can strip the spaces
screenshot 2019-03-05 at 17 28 32

nakatuddesuzan: This means, I need to refactor `validate empty fields`. Thanks for the pointer let me work on it. I guess this is a bug to fix on V1 too. cc @Bonifase , cc @joshuaocero

tests/test_office_structure/test_add_node.py
```diff
@@ -0,0 +1,43 @@
...
+        """
+        CommonTestCases.admin_token_assert_in(
+            self,
+            duplicate_node_creation,
+            "location node level already exists"
+        )
+
+    def test_add_child_node_with_non_existent_parent_id(self):
+        """
+        Tests whether an admin can create
```
Bonifase: We are not testing whether this can happen, we are testing that this cannot happen

tests/test_office_structure/test_add_node.py
```diff
@@ -0,0 +1,43 @@
...
+
+        CommonTestCases.admin_token_assert_in(
+            self,
+            node_creation_mutation,
+            "floors"
+        )
+
+    def test_add_duplicate_node(self):
+        """
+        Tests whether an admin can create two nodes
```
Bonifase: We are not testing whether this can happen, we are testing that this cannot happen

api/office_structure/models.py
```diff
@@ -17,4 +19,18 @@ class OfficeStructure(Base, Utility, BaseNestedSets):
     )
 
     def __repr__(self):
-        return "<{}>".format(self.name)
+        return "<OfficeStructure {}>".format(self.name)
+
+    def add_node(self, name, parent_id=None):
```
bevkololi: This function has not been tested according to circle-ci. Is there a way you could test this?

api/office_structure/models.py
```diff
@@ -17,4 +19,18 @@ class OfficeStructure(Base, Utility, BaseNestedSets):
...
+        """Function for adding nodes"""
+        if parent_id is None:
+            db_session.add(OfficeStructure(name=name))
+        else:
+            db_session.add(OfficeStructure(name=name, parent_id=parent_id))
+
+    def add_branch(self, nodes):
```
bevkololi: Same as this function

api/office_structure/schema.py
nakatuddesuzan: This `https://www.pivotaltracker.com/story/show/164361872` on the PT board would solve the issue. Person working on that task can use the same solution I have used here for v1 only, v2 will use this once this PR is merged. This has been solved @michael-basweti

tests/test_office_structure/test_add_node.py
nakatuddesuzan: Thanks @Bonifase , This has been worked on

tests/test_office_structure/test_add_node.py
nakatuddesuzan: This has been modified @Bonifase

tests/test_office_structure/test_add_node.py
```diff
@@ -0,0 +1,43 @@
...
+    duplicate_node_creation,
+    child_node_creation_mutation_with_non_existent_parent_id
+)
+
+
+class TestAddNode(BaseTestCase):
+
+    def test_add_node(self):
+        """
+        Tests an admin can node create a node
```
Bonifase: Kindly pay more attention to your typo errors

utilities/validations.py
```diff
@@ -66,3 +70,14 @@ def validate_timezone_field(**kwargs):
...
     if time_zone not in timezones:
         raise AttributeError("Not a valid time zone")
+
+
+def validate_parent_node_id(parent_id):
+    """
+    Function to validate Parent node ID
+    when creating child node
+    """
+    if parent_id:
```
Bonifase: What happens if the `node_id` is not provided?

nakatuddesuzan: If a node_id(parent_id) is not provided, that means a new root has been created hence a new tree

tests/test_office_structure/test_add_node.py
nakatuddesuzan: Sorry for this. This has been fixed

utilities/validations.py
nakatuddesuzan: so what am doing here is trying to validate the parent_id once provided. The validation is about preventing the creation of a node with a parent_id that doesn't exist.

kampkelly:
The mutation works but there are some concerns which have already been pointed out such as the creation of a node with an empty string by @michael-basweti .
 Please look into it and others @nakatuddesuzan

nakatuddesuzan:
> The mutation works but there are some concerns which have already been pointed out such as the creation of a node with an empty string by @michael-basweti .
> Please look into it and others @nakatuddesuzan

Hey @kampkelly ,  @michael-basweti's  issue was already resolved, kindly test it out and see if it's working on your end. If not you will need to pull my branch again to get the new changes

utilities/validations.py
nakatuddesuzan: One has to always provide a parent_node when creating a child for that parent

kampkelly:
> > The mutation works but there are some concerns which have already been pointed out such as the creation of a node with an empty string by @michael-basweti .
> > Please look into it and others @nakatuddesuzan
> 
> Hey @kampkelly , @michael-basweti's issue was already resolved, kindly test it out and see if it's working on your end. If not you will need to pull my branch again to get the new changes

Pulled your branch and I have seen it has been fixed.

api/office_structure/models.py
nakatuddesuzan: Yes @bevkololi , there is away and I have added a test for them. Thanks for the pointer

tests/test_office_structure/test_add_node.py
```diff
@@ -0,0 +1,81 @@
...
+        structure.add_node(name="root")
+        # Add root node
+        root_node = OfficeStructure.query.filter_by(id=1).first()
+        self.assertEqual(root_node.parent_id, None)
+        # Add child node
+        structure.add_node(name="floor", parent_id=1)
+        child_node = OfficeStructure.query.filter_by(id=2).first()
+        self.assertEqual(child_node.parent_id, 1)
+        nodes = OfficeStructure.query.all()
+        self.assertEqual(len(nodes), 4)
```
nakatuddesuzan: Nb: On this line am including the two nodes that are in the base.py that are being used to test the graphl endpoint. Am adding two nodes here to the two in the `base.py` file. this is why am asserting Equal to 4

tests/test_office_structure/test_add_node.py
```diff
@@ -0,0 +1,81 @@
...
+        structure.add_node(name="root")
+        # create list of nodes for branch
+        nodes = []
+        node_1 = OfficeStructure(name="floor", parent_id=1)
+        nodes.append(node_1)
+        node_2 = OfficeStructure(name="room", parent_id=2)
+        nodes.append(node_2)
+        structure.add_branch(nodes)
+        nodes_list = OfficeStructure.query.all()
+        self.assertEqual(len(nodes_list), 5)
```
nakatuddesuzan: Nb: same applies her `2+3`

utilities/validations.py
nakatuddesuzan: Hey @Bonifase , I am not validating this in the schema. This check is necessary because if I don't include it then one will just create any node and provide a parent_id that does not exist. Even though the parent_id field of that child won't be populated, I think it is necessary to warn the user that the parent_id he or she is trying to use does not exist.

Bonifase: > Hey @Bonifase , I am not validating this in the schema. This check is necessary because if I don't include it then one will just create any node and provide a parent_id that does not exist. Even though the parent_id field of that child won't be populated, I think it is necessary to warn the user that the parent_id he or she is trying to use does not exist.

My concern is why use the `if parent_id:` statement yet it has to be provided always. Why can't you just do the query in line 79 without line 78?

dnuwa: Awesome work @nakatuddesuzan this `if parent_id:` is redundant thus doing away with it, in this case, is a good option.

nakatuddesuzan: @Bonifase and @dnuwa  , That line is necessary `parent_id` is not always providedfor examplee when creating a root not(That is when starting a new tree). Parent_ids are only provided when creating child nodes. So that `if statement` is checking when for when the parent_id is provided

nakatuddesuzan: For clarity here, because the root is created without a parent_id,  if i remove the check ('if parent_id'). I won't be able to create that root as it will be checking for a parent_id and yet it is not provided. You can try this out your end. Remove the if statement and run this mutation, you will prove what am taking about and understand it better
```
mutation{
  createNode(name:"root"){
    node{
      name
      level
      left
      right
      parentId
    }
    
  }
}
```

dnuwa:
Works as explained @nakatuddesuzan

utilities/validations.py
nakatuddesuzan: Hello @Bonifase , Thanks for the pointers I have more made it more readable like advised

PR #291, opened 2019-03-05
#164266587 An admin should be able to add structure id to room model

RachaelNantale:
#### Description
Currently, there is no connection between the structure table and the Room table. This Pr adds that connection by adding a `structure_id` to the Room Model.

#### Type of change
- New feature (non-breaking change which adds functionality)

- Pull this branch `story/164266587-add-strucutre-id-to-room-model`
- Run the `createRoom` mutation
- You should be able to create a room with a structureId.

#### Checklist:

#### Pivotal Tracker
[#164266587](https://www.pivotaltracker.com/story/show/164266587)


api/room/models.py
```diff
@@ -12,6 +12,7 @@
 from api.tag.models import Tag  # noqa: F401
 from utilities.validator import verify_calendar_id
 from api.devices.models import Devices # noqa F4
+from api.office_structure.models import OfficeStructure  # noqa f401
```
kampkelly: You can remove this noqa here.
Also, this import is not being used in this file, that is why you may be getting unused import.

api/room/models.py
```diff
@@ -36,12 +37,19 @@ class Room(Base, Utility):
         ForeignKey('locations.id', ondelete="CASCADE"),
         nullable=True
     )
+    structure_id = Column(
+        Integer,
+        ForeignKey('structure.id', ondelete="CASCADE"),
+        nullable=True   # TODO: Change this to False
+                        # after migrating from v1 to v2
+    )
```
kampkelly: To clarify, is this to prevent a migration error due to data already present in the tags table?

RachaelNantale: No in the `room` table itself. There is already existing data in the room table.

api/room/models.py
RachaelNantale: Thanks, @kampkelly, This was failing before without the import but it has been removed.

api/room/models.py
kampkelly: Okay.

dnuwa:
@RachaelNantale  kindly update the checklist so the team can be guided while reviewing.

api/room/models.py
```diff
@@ -36,12 +36,19 @@ class Room(Base, Utility):
         ForeignKey('locations.id', ondelete="CASCADE"),
         nullable=True
     )
+    structure_id = Column(
+        Integer,
+        ForeignKey('structure.id', ondelete="CASCADE"),
+        nullable=True   # TODO: Change this to False
```
ThaDeveloper: Great work @RachaelNantale, but should `nullable` not be `False`?

RachaelNantale: Thanks, @ThaDeveloper,  as advised by @joshuaocero,  I used `True` because of the already existing data in the database would not be lost if my migration is run. As explained in the `TODO` comment.  This will be changed to False after the migration from v1 to v2.

api/room/schema.py
```diff
@@ -67,6 +68,7 @@ class Arguments:
         image_url = graphene.String()
         location_id = graphene.Int(required=True)
         calendar_id = graphene.String()
+        structure_id = graphene.Int()
```
Bonifase: Don't you think is in necessary to make this field a required field?

api/room/schema.py
```diff
@@ -93,8 +95,15 @@ def mutate(self, info, **kwargs):
         if kwargs.get('room_tags'):
             room_tags = kwargs.pop('room_tags')
         room = RoomModel(**kwargs)
-        save_room_tags(room, room_tags)
-        return CreateRoom(room=room)
+        payload = {
```
Bonifase: What happens when the user fails to provide `structure_id`?

api/room/schema.py
RachaelNantale: Thanks,  @Bonifase  for pointing this out.  I have changed it.

api/room/schema.py
RachaelNantale: The user gets an error message telling them that the `structure_id ` is needed.

dnuwa:
well done @RachaelNantale I am curious though, what happens when one provides a `structure_id` that is non-existing in the `structure` table?

RachaelNantale:
> well done @RachaelNantale I am curious though, what happens when one provides a `structure_id` that is non-existing in the `structure` table?

The user gets an error message that the `structure_id` doesn't exist

Bonifase:
This PR is on an old office structure which is no longer being used.

PR #303, opened 2019-03-11
#164361872 Validate Questions Creation.

verenceLola:
#### Description
Currently, you can create a question with empty fields with just spaces. The PR strips these spaces so you don't create a question with empty fields

#### Type of change
Bug (validate question creation)

 - Checkout by running `git checkout bg-validate-questions-creation-164361872`
 - Perform the mutation below with empty `quetionTitle`:
```
mutation {
  createQuestion(questionType:"1", questionTitle:"   ", question:"How would you rate the cleanliness of this room",
    startDate: "2019-03-12T14:25:00", endDate: "2019-03-19T14:25:37")
{
  question{
    id,
    questionTitle,
    questionType,
    question
  }
}
}
```
 - create Question with all the required:
```
mutation {
  createQuestion(questionType:"Ratings", questionTitle:"Room Tablets", question:"How would you rate the cleanliness of this room",
    startDate: "2019-03-12T14:25:00", endDate: "2019-03-19T14:25:37")
{
  question{
    id,
    questionTitle,
    questionType,
    question
  }
}
}
```
The mutation above create a new Question.

[#164361872](https://www.pivotaltracker.com/story/show/164361872)

#### Checklist:

### Screenshots
 - ## Before
 - #### create Question with empty field.
image

 - ## After
 - #### create Question with empty field.
image

### DIFF
--- a/utilities/validations.py
+++ b/utilities/validations.py
@@ -22,8 +22,20 @@ def validate_empty_fields(**kwargs):
     """
     for field in kwargs:
         value = kwargs.get(field)
+        if isinstance(value, str):
+            value = value.strip()
         if not type(value) is bool and not value:
             raise AttributeError(field + " is required field")
+
+
+def validate_question_length(question_title):
+    """
+    Function to validate the question length is less than
+    20 characters
+    :params question
+    """
+    if not len(question_title) < 20:
+        raise AttributeError("questionTitle should be less that 20 characters")
 
 
 def validate_date_time_range(**kwargs):
--- a/api/question/schema.py
+++ b/api/question/schema.py
@@ -6,7 +6,8 @@ from api.question.models import Question as QuestionModel
 from utilities.validations import (
     validate_empty_fields,
     validate_date_time_range,
-    validate_question_type
+    validate_question_type,
+    validate_question_length
     )
 from utilities.utility import update_entity_fields
 from helpers.auth.authentication import Auth
@@ -44,6 +45,7 @@ class CreateQuestion(graphene.Mutation):
     def mutate(self, info, **kwargs):
         validate_empty_fields(**kwargs)
         validate_question_type(**kwargs)
+        validate_question_length(kwargs['question_title'])
         validate_date_time_range(**kwargs)
         question = QuestionModel(**kwargs)
         payload = {
--- a/fixtures/questions/create_questions_fixtures.py
+++ b/fixtures/questions/create_questions_fixtures.py
@@ -53,6 +53,24 @@ create_question_response = {
   }
 }
 
+question_mutation_with_very_long_title = '''
+     mutation{{
+  createQuestion(questionType:"Rate",
+  questionTitle:"This is a very very long question title"
+  question:"How will you rate the brightness of the room",
+  startDate:"{}", endDate:"{}") {{
+    question{{
+      id
+      question
+      questionType
+      startDate
+      endDate
+      }}
+  }}
+}}
+'''.format(start_date, end_date)
+
+
 question_mutation_query_without_name = '''
      mutation{{
   createQuestion(questionType:"",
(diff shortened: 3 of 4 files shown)


utilities/validations.py
```diff
@@ -24,6 +24,8 @@ def validate_empty_fields(**kwargs):
         value = kwargs.get(field)
         if not type(value) is bool and not value:
             raise AttributeError(field + " is required field")
+        if isinstance(value, str) and not value.strip():  # remove white space
```
nakatuddesuzan: Hello @verenceLola, great work here. Kindly remove the spaces at the ends too if that is possible.

utilities/validations.py
```diff
@@ -24,6 +24,8 @@ def validate_empty_fields(**kwargs):
...
+            raise AttributeError(field + " cannot be empty")
```
nakatuddesuzan: The message `can not be empty` and `required field`  Kind of look similar to me. We could use a better message here or use the same message for both by refactoring this code like this:
```
for field in kwargs:
        value = kwargs.get(field)
        if isinstance(value, str):
            value = value.strip()
        if not type(value) is bool and not value:
            raise AttributeError(field + " is required field")
```

nakatuddesuzan: Kindly fix the code climate issue too

utilities/validations.py
verenceLola: @nakatuddesuzan I'm not sure what you mean here. Kindly shade more colour on this. Thanks

utilities/validations.py
verenceLola: @nakatuddesuzan`required field` validates the existence of a field. On the other hand, empty validates if it has a value i.e it doesn't contain white spaces only. Note: a field can exist and be empty  but cannot be non-existent and be empty.

utilities/validations.py
nakatuddesuzan: Screen Shot 2019-03-11 at 21 30 49

Screen Shot 2019-03-11 at 21 31 12
checkout those screenshots. The same question is created, the difference is that the second has spaces at the end. Try it out and see what am talking about.

utilities/validations.py
nakatuddesuzan: Hello @verenceLola ,  I get the point.  how about we use `Spaces are not allowed!` for the `RegularExpressionValidator` and `field cannot be empty` or `field is required` for `RequiredFieldValidator`. The latter 2 messages mean the same.

nakatuddesuzan: Hey @verenceLola, The message `cannot be empty` though I still feel like those two instances should display the same message since the validation is similar. Empty space means there is no string in the field hence that field is required.

ThaDeveloper: I think @nakatuddesuzan suggestion is better, `""` and `"  "` should get the same message.

dnuwa: good work @verenceLola though I request that we have the same message as mentioned above.

verenceLola:
> Good work @verenceLola Implement the requested changes by @nakatuddesuzan and it will be good.

I have implemented the requested changes. Thanks for the feedback @TeamoreA , @nakatuddesuzan

utilities/validations.py
```diff
@@ -22,6 +22,8 @@ def validate_empty_fields(**kwargs):
     """
     for field in kwargs:
         value = kwargs.get(field)
+        if isinstance(value, str):
+            value = value.strip()  # remove white space
```
dnuwa: awesome work @verenceLola though I think we should do away with the inline comment.

verenceLola: Thanks @dnuwa. Let me work on that.

dnuwa:
tested and works as described on the PT board.

verenceLola:
> Kindly update the question title length to 20 instead of 16. This is a request from the android team

@Bonifase This has been worked on.

PR #304, opened 2019-03-11
#164457411 An admin should be able to get structure information when querying locations

RachaelNantale:
#### Description
Currently, When querying locations,  passing in children in GraphQL only brings back one level. This PR brings back all the  nodes or levels for that location
#### Type of change
- New feature (non-breaking change which adds functionality)

- Pull this branch `story/164457411-add-structure-info-to-getLocation-query`
- Run the `allLocations` query with 
```
{
    allLocations{
        name
        abbreviation
      	structureId
        structure{
            name
            getNodes{
              id
              name
              level
            }
        }
    }
}
```
- You should be able to view all the nodes associated with that location.

#### Checklist:

#### Pivotal Tracker
[#164457411](https://www.pivotaltracker.com/story/show/164457411)
#### Screenshots if any
Screen Shot 2019-03-18 at 2 04 12 PM

#### Any background context that you would like to provide:
- The data returned are all the nodes associated with that location.


api/office_structure/models.py
```diff
@@ -38,3 +38,27 @@ def add_branch(self, nodes):
         db_session.add_all(nodes)
 
         db_session.commit()
+
+    @property
+    def nested_children(self):
+        """
+        Function to return a dictionary of the nested children
```
nakatuddesuzan: Hello @RachaelNantale , good work.  I think this function returns a list, not a dictionary. Kindly confirm this. If it's true, please update the docstring.

nakatuddesuzan: To add on, Is it possible to write tests for these functions?

RachaelNantale: Thanks, @nakatuddesuzan  for the feedback. so for your first question. This returns a dictionary

nakatuddesuzan: Okay, I hadn't noticed the use `retrieve_nested_data` function here.

RachaelNantale:
> What is the rationale of returning nested parents as a string instead of an object?
The type of data being returned is dynamic and so the only way we can return data whose size (children) we do not know is through the use of a JSON string.  The graphene types that return objects are `graphene.Int` and `graphene.String` which is not the type of data being returned.

api/office_structure/models.py
RachaelNantale: hey @nakatuddesuzan, I have realised that since this function is being used when querying all locations, it has already been covered.

dnuwa:
Good work @RachaelNantale this has been tested and works as needed.

RachaelNantale:
> What is the rationale of returning nested parents as a string instead of an object

> > What is the rationale of returning nested parents as a string instead of an object?
> > The type of data being returned is dynamic and so the only way we can return data whose size (children) we do not know is through the use of a JSON string.  The graphene types that return objects are `graphene.Int` and `graphene.String` which is not the type of data being returned.

This implementation has been changed. The data returned now is a list of all the nodes connected to the associated location.

api/office_structure/models.py
```diff
@@ -38,3 +38,11 @@ def add_branch(self, nodes):
         db_session.add_all(nodes)
 
         db_session.commit()
+
+    @property
+    def get_nodes(self):
```
ThaDeveloper: Good work @RachaelNantale, I think this property/field is better referred to as `nodes`.

RachaelNantale: Thanks @ThaDeveloper, as advised by @Bonifase,  a function's name should begin with a verb

RachaelNantale:
> Add endpoint documentation to this endpoint

This has been implemented

api/office_structure/models.py
ThaDeveloper: Okay,my concern was this ```structure{
            name
            getNodes{
              id
              name
              level
            }```. Whether we could have getNodes as `nodes`

RachaelNantale: Yes, this gets its name from the `get_nodes` function. Whichever name I use here is what I have to call the field.

Bonifase:
This PR is on an office structure which is no longer being used

PR #316, opened 2019-03-20
#164657362 Refactor get questions endpoint

ThaDeveloper:
#### Description
Refactor get all questions endpoint to only return valid/active questions.

#### Type of change
Chore (Refactor get all questions)

- git pull and checkout branch `ch-refactor-get-questions-164657362`
- run the query below.
````
  query {
    allQuestions(startDate: "2018 Nov 20", endDate: "2018 Nov 28") {
      id
      questionType
      questionTitle
      question
      startDate
      endDate
      questionResponseCount
      response {
        id
        questionId
        roomId
      }
  }
}
````

#### Checklist:

[#164657362](https://www.pivotaltracker.com/story/show/164657362)

### DIFF
--- a/helpers/questions_filter/questions_filter.py
+++ b/helpers/questions_filter/questions_filter.py
@@ -1,0 +1,33 @@
+from dateutil import parser
+from dateutil.relativedelta import relativedelta
+from graphql import GraphQLError
+
+
+def filter_questions_by_date_range(questions, start_date, end_date):
+    """
+    Return questions that  fall in the date range
+    """
+    if not (start_date and end_date):
+        return questions
+    start_date, end_date = format_range_dates(start_date, end_date)
+    filtered_questions = []
+    for question in questions:
+        question_start_date = parser.parse(question.start_date)
+        question_end_date = parser.parse(question.end_date)
+        if question_start_date >= start_date and question_end_date <= end_date:
+            filtered_questions.append(question)
+    return filtered_questions
+
+
+def format_range_dates(start_date, end_date):
+    """
+    Convert dates to date objects and add one day to end_date
+    Data from front-end doesn't include time
+    """
+    start_date = parser.parse(start_date).strftime('%Y-%m-%d')
+    end_date = parser.parse(end_date).strftime('%Y-%m-%d')
+    if start_date > end_date:
+        raise GraphQLError("Start date must be lower than end date")
+    start_date = parser.parse(start_date)
+    end_date = parser.parse(end_date) + relativedelta(days=1)
+    return(start_date, end_date)
--- a/api/question/schema.py
+++ b/api/question/schema.py
@@ -11,6 +11,9 @@ from utilities.utility import update_entity_fields
 from helpers.auth.authentication import Auth
 from helpers.auth.error_handler import SaveContextManager
 from helpers.pagination.paginate import Paginate, validate_page
+from helpers.questions_filter.questions_filter import (
+    filter_questions_by_date_range
+)
 
 
 class Question(SQLAlchemyObjectType):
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: helpers/questions_filter/questions_filter.py [hunk 1]
```diff
@@ -0,0 +1,33 @@
...
+    questions = []
+    for question in query:
+        question_start_date = parser.parse(question.start_date)
+        question_end_date = parser.parse(question.end_date)
+        if question_start_date >= start_date and question_end_date <= end_date:
+            questions.append(question)
+    return questions
+
+
+def format_range_dates(start_date, end_date):
```
RachaelNantale: hello @ThaDeveloper, there is a function in the codebase called `validate_date_time_range` in `utitlities\validations.py` that does the exact implementation. Can we use that rather than creating another one

ThaDeveloper: @RachaelNantale The reason `validate_date_time_range` won't be suited in this case is that it only does validation which isn't much applicable when filtering questions, for instance, the condition below only applies when creating questions but we can get questions whose start_date is lower than today.  Also, `format_range_dates` convert the date format based on how we are storing data in the backend.

RachaelNantale: @ThaDeveloper , so these functions are basically doing the same thing.  The one already in the codebase can changed to cater for  this as seen below 
```def validate_date_time_range(**kwargs):
    """
    Function to validate the dates entered
    for questions for feedback
    :params kwargs
    """
    if 'start_date' in kwargs and isinstance(kwargs['start_date'], str):
        kwargs['start_date'] = datetime.datetime.strptime(kwargs['start_date'], '%Y %b %d')
        kwargs['end_date'] = datetime.datetime.strptime(kwargs['end_date'], '%Y %b %d')
    if ('start_date' and 'end_date' in kwargs) and\
            kwargs['start_date'] < datetime.datetime.now():
        raise ValueError('startDate should be today or after')
    elif ('start_date' and 'end_date' in kwargs) and\
            kwargs['end_date'] - kwargs['start_date'] < datetime.timedelta(
                days=1
            ):
        raise ValueError(
            'endDate should be at least a day after startDate'
        )
```

nakatuddesuzan: Hello @ThaDeveloper , If you need more light on the issue raised by @RachaelNantale , Kindly ping me for assistance. Otherwise, that issue raised is valid.

ThaDeveloper: @nakatuddesuzan I just had sync with her, there's something she's confirming. Thanks

helpers/questions_filter/questions_filter.py
```diff
@@ -0,0 +1,33 @@
+from dateutil import parser
+from dateutil.relativedelta import relativedelta
+from graphql import GraphQLError
+
+
+def filter_questions_by_date_range(query, start_date, end_date):
```
Bonifase: Kindly change query to questions, same to line 14

helpers/questions_filter/questions_filter.py
```diff
@@ -0,0 +1,33 @@
...
+    """
+    Return questions that  fall in the date range
+    """
+    if not (start_date and end_date):
+        return query
+    start_date, end_date = format_range_dates(start_date, end_date)
+    questions = []
```
Bonifase: Chenge this to filtered_questions

helpers/questions_filter/questions_filter.py
ThaDeveloper: @Bonifase This has been changed.

dnuwa:
good work @ThaDeveloper tested and works as expected.

PR #337, opened 2019-04-02
#164471444 Fetch Total Booking Count for a Specific Room (V2) 

dnuwa:
#### What does this PR do?
Fetch Total Booking Count for a Specific Room
Ensure that the user can view the booking count of a single room.
git checkout `ft-room-booking-count-164471444`
run the mutation below to view bookings for a room with id `1`
Daily bookings count
````
query {
  bookingsAnalyticsCount(startDate:"Nov 10 2018", endDate:"Nov 12 2018", roomId:1){
   roomName
   period
   bookings
  }
}
````
monthly room bookings query
````
query {
  bookingsAnalyticsCount(startDate:"Aug 1 2018", endDate:"Nov 1 2018", roomId:4){
    roomName
   period
   bookings
  }
}
````
[#164471444](https://www.pivotaltracker.com/story/show/164471444)
#### Screenshots
Daily bookings count
image
Monthly bookings count
image

### DIFF
--- a/api/room/schema_query.py
+++ b/api/room/schema_query.py
@@ -261,6 +261,7 @@ class Query(graphene.ObjectType):
         BookingsAnalyticsCount,
         start_date=graphene.String(required=True),
         end_date=graphene.String(required=True),
+        room_id=graphene.Int(),
         description="Returns the total number of room bookings and accepts the arguments\
             \n- start_date: Start date when you want to get analytics from\
             [required]\n- end_date: The end date to take the analytics upto\
@@ -411,11 +412,13 @@ class Query(graphene.ObjectType):
 
     @Auth.user_roles('Admin', 'Default User')
     def resolve_bookings_analytics_count(
-            self, info, start_date, end_date):
-        # Getting booking analytics count
+            self, info, **kwargs):
+        start_date, end_date, room_id = (kwargs.get('start_date'),
+                                         kwargs.get('end_date'),
+                                         kwargs.get('room_id'))
         query = Room.get_query(info)
         analytics = RoomAnalyticsRatios.get_bookings_analytics_count(
-            self, query, start_date, end_date)
+            self, query, start_date, end_date, room_id=room_id)
         return analytics
 
     def resolve_analytics_for_daily_room_events(
--- a/helpers/calendar/analytics_helper.py
+++ b/helpers/calendar/analytics_helper.py
@@ -4,6 +4,7 @@ import dateutil.parser
 from dateutil.relativedelta import relativedelta
 from graphql import GraphQLError
 from collections import Counter
+from api.room.models import Room as RoomModel
 from api.location.models import Location as LocationModel
 from helpers.room_filter.room_filter import room_join_location
 from helpers.auth.admin_roles import admin_roles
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

File: api/room/schema_query.py [hunk 1]
```diff
@@ -331,10 +332,10 @@ def resolve_analytics_ratios_per_room(
 
     @Auth.user_roles('Admin', 'Default User')
     def resolve_bookings_analytics_count(
-            self, info, start_date, end_date):
+            self, info, start_date, end_date, room_id=None):
```
verenceLola: Don't you think it's better to use `kwargs` and leave only `self`, and `info` as the positional arguments?

helpers/calendar/analytics_helper.py
```diff
@@ -175,16 +176,37 @@ def get_room_statistics(self, number_of_events_in_room, all_details):
...
-                self, room["room_id"], start_date, end_date)
-            if all_events:
-                bookings += len(all_events)
+            if room_id:
+                exact_room = active_rooms.filter(
+                    RoomModel.id == room_id).first()
+                if not exact_room:
+                    raise GraphQLError("Room Id does not exist")
+                if room.get('calendar_id') == exact_room.calendar_id:
+                    bookings = CommonAnalytics.get_bookings_count(instance,
```
verenceLola: I believe this section can be formatted better, those are a lot of unnecessary lines wasted. What do you think?

helpers/calendar/analytics_helper.py
```diff
@@ -175,16 +176,37 @@ def get_room_statistics(self, number_of_events_in_room, all_details):
...
-            all_events = CommonAnalytics.get_all_events_in_a_room(
-                self, room["room_id"], start_date, end_date)
-            if all_events:
-                bookings += len(all_events)
+            if room_id:
+                exact_room = active_rooms.filter(
+                    RoomModel.id == room_id).first()
+                if not exact_room:
+                    raise GraphQLError("Room Id does not exist")
+                if room.get('calendar_id') == exact_room.calendar_id:
```
Bonifase: I don't think this line is necessary here.

dnuwa: @Bonifase this condition enables one to check for a specific calendar_Id whose booking count we need to be returned. so believe it is necessary. kindly shed some more light on the same if you think otherwise.

api/room/schema_query.py
dnuwa: thanx @verenceLola this has been worked on.

helpers/calendar/analytics_helper.py
dnuwa: Well, I think these lines are necessary because first we have to cater for a scenario where the room Id passed was wrong, then cater for a scenario where a correct Id was passed and hence the results required pertain to a single room. After this, we then proceed to cater for a scenario where no id was passed and hence results returned pertain to all rooms. How do you think it could be made better while maintaining this functionality? @verenceLola

verenceLola: something like this, 
```
bookings = CommonAnalytics.get_bookings_count(
               instance, room, start_date, end_date, bookings
           )
```

helpers/calendar/analytics_helper.py
```diff
@@ -177,16 +178,37 @@ def get_room_statistics(self, number_of_events_in_room, all_details):
...
+                    bookings = CommonAnalytics.get_bookings_count(instance,
+                                                                  room,
+                                                                  start_date,
+                                                                  end_date,
+                                                                  bookings)
+            else:
```
ThaDeveloper: @dnuwa I think you can get rid of this else statement and still have same functionality

helpers/calendar/analytics_helper.py
dnuwa: thanx @verenceLola though with this, there is no way to confirm that a particular `calendar_id` actually exists. this would work if this function was only returning bookings for all rooms in a particular location which already does though it was simply modified to also take care for booking count for a single room in a location. lets us sync so I can try to clarify more.

helpers/calendar/analytics_helper.py
dnuwa: @ThaDeveloper when the `else` is removed each time this function is run, we'll have wrong data for the booking count of the single room because it will always return the bookings count for all the rooms. kindly try it out, you will see what I mean.

ThaDeveloper: @dnuwa I had tried it like below and got similar results
image

helpers/calendar/analytics_helper.py
dnuwa: thanx @verenceLola this change has been implemented

helpers/calendar/analytics_helper.py
dnuwa: this was fixed. thanx @ThaDeveloper

helpers/calendar/analytics_helper.py
dnuwa: thanx for point this out @Bonifase. this has been refactored.

PR #346, opened 2019-04-05
#165124861 Add User Location Automatically upon User Registration

verenceLola:
### What does this PR do?
Add user location automatically upon user registration 
Upon successful user registration, update the location table to add the location of the user if it doesn't exist in the database.
 - Checkout by running `git checkout ft-add-location-name-165124861`
 - Perform any query or mutation with a valid token from a new user(not already in the database)
 - Check your users table to make sure that the user was successfully added. Now open the location table to confirm that the new user location has also been added.

Currently, when a new user is added into the database, the location of the user has to be manually added to the location table. To prevent this, upon user registration, the location of the user should be added automatically into the database.
Since the user information only has the location name and not other fields present in the location table, constraints such as NOT NULL should be removed to permit NULL values in fields such as `abbreviation`. 
[#165124861](https://www.pivotaltracker.com/story/show/165124861)


api/location/models.py
```diff
@@ -23,15 +23,14 @@ class Location(Base, Utility):
     id = Column(Integer, Sequence('locations_id_seq',
                                   start=1, increment=1), primary_key=True)
     name = Column(String, nullable=False)
-    abbreviation = Column(String, nullable=False)
```
RachaelNantale: Great work @verenceLola, why did you remove this?
same as line 37

verenceLola: Kindly have a look at the `background context` of the PR description. I've updated the information.

helpers/auth/authentication.py
```diff
@@ -135,9 +136,15 @@ def wrapper(*args, **kwargs):
                             if value['location']:
                                 user.location = value['location']['name']
                                 user.save()
+                                add_location(
+                                    value['location']['name']
+                                )  # add location name if not in database
                             else:
                                 user.location = "Nairobi"
```
KMaina: Good work, why are we defaulting to Nairobi here?

helpers/auth/authentication.py
```diff
@@ -129,15 +125,25 @@ def wrapper(*args, **kwargs):
...
                                 user.save()
+                                add_location(
+                                    "Nairobi"
```
TeamoreA: Good work @verenceLola . However, I didn't get the reason behind defaulting the location to Nairobi yet your PR should aim at saving the user's exact location.

verenceLola: Still, if the location of the user cannot be fetched from  the Andela API, it's defaulted to `Nairobi` as before. @TeamoreA

TeamoreA: Alright👍🏼

joshuaocero: @verenceLola take this opportunity to refactor the duplication in the `if..else` statement. Remove the `else`. Can we also make the `add_location` function name more descriptive? Because it does not only add the location, it also performs a check. Otherwise well done!

verenceLola:
> Kindly check the structure of your commit message.

Thanks for the feedback @Bonifase .This has been rectified

PR #341, opened 2019-04-03
#164890924 Add mutation to end meeting

ThaDeveloper:
#### What does this PR do?
Add a mutation to end meeting
#### Description of Task to be completed?
As a user I want to end a meeting from the mobile app and have this updated in the event table so that I can save the time when the meeting actually ended
#### How should this be manually tested?
- git pull and checkout branch ft-end-meeting-mutation-164890924
- run application
- Ensure you have a checked-in event
- run the mutation below
```
mutation {
 endEvent(calendarId:"andela.com_353633323439323830@resource.calendar.google.com",
        eventId:"3f85pi6jgpc1qslnp312k3k4fl", startTime:"2018-04-10T01:00:00-07:00",
        endTime:"2018-04-10T03:00:00-07:00", meetingEndTime:"2018-02-07T04:15:00-08:00")
        {
            event{
                eventId
              	eventTitle
                checkedIn
                cancelled
              	meetingEndTime
                room{
                    id
                    name
                    calendarId
                }
            }
        }
    }

```
#### Checklist:
- [x]  My code follows the style guidelines of this project
- [x] I have linted my code prior to submission
- [x] I have made the corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my feature works
- [x] New and existing unit tests pass locally with my changes

#### What are the relevant pivotal tracker stories?
[#164890924](https://www.pivotaltracker.com/story/show/164890924)

PR #356, opened 2019-04-11
#164890924 Add mutation to end meeting v1

ThaDeveloper:
#### What does this PR do?
Add a mutation to end meeting. This PR updates `develop` with the changes introduced to `v2` [here](https://github.com/andela/mrm_api/pull/341).
As a user I want to end a meeting from the mobile app and have this updated in the event table so that I can save the time when the meeting actually ended
- git pull and checkout branch ft-end-meeting-mutation-v1-164890924
- run application
- Ensure you have a checked-in event
- run the mutation below
```
mutation {
 endEvent(calendarId:"andela.com_353633323439323830@resource.calendar.google.com",
        eventId:"3f85pi6jgpc1qslnp312k3k4fl", startTime:"2018-04-10T01:00:00-07:00",
        endTime:"2018-04-10T03:00:00-07:00", meetingEndTime:"2018-02-07T04:15:00-08:00")
        {
            event{
                eventId
              	eventTitle
                checkedIn
                cancelled
              	meetingEndTime
                room{
                    id
                    name
                    calendarId
                }
            }
        }
    }

```
#### Checklist:

[#164890924](https://www.pivotaltracker.com/story/show/164890924)

### DIFF
--- a/api/events/schema.py
+++ b/api/events/schema.py
@@ -1,5 +1,6 @@
 import graphene
 from graphene_sqlalchemy import SQLAlchemyObjectType
+from graphql import GraphQLError
 
 from api.events.models import Events as EventsModel
 from helpers.calendar.events import RoomSchedules
@@ -69,6 +70,31 @@ class CancelEvent(graphene.Mutation):
         return CancelEvent(event=event)
 
 
+class EndEvent(graphene.Mutation):
+    """
+    Mutation to end an event
+    Returns event payload on ending the event
+    """
+    class Arguments:
+        calendar_id = graphene.String(required=True)
+        event_id = graphene.String(required=True)
+        start_time = graphene.String(required=True)
+        end_time = graphene.String(required=True)
+        meeting_end_time = graphene.String(required=True)
+    event = graphene.Field(Events)
+
+    def mutate(self, info, **kwargs):
+        room_id, event = check_event_in_db(self, info, "ended", **kwargs)
+        if not event:
+            event = EventsModel(
+                event_id=kwargs['event_id'],
+                meeting_end_time=kwargs['meeting_end_time']
+                )
+            event.save()
+
+        return EndEvent(event=event)
+
+
 def check_event_in_db(instance, info, event_check, **kwargs):
     room_id = RoomSchedules().check_event_status(info, **kwargs)
     event = EventsModel.query.filter_by(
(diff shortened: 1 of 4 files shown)


api/events/schema.py
```diff
@@ -69,6 +70,31 @@ def mutate(self, info, **kwargs):
...
+
+    def mutate(self, info, **kwargs):
+        room_id, event = check_event_in_db(self, info, "ended", **kwargs)
+        if not event:
+            event = EventsModel(
+                event_id=kwargs['event_id'],
+                meeting_end_time=kwargs['meeting_end_time']
+                )
+            event.save()
+
```
rajeman: @ThaDeveloper 
Why you are creating the event if it is not in the db? 
since what is required is to end the event.

api/events/schema.py
```diff
@@ -104,3 +136,13 @@ class Mutation(graphene.ObjectType):
             [required]\n- start_time: The start time of the calendar event\
             [required]\n- end_time: The field with the end time of the calendar\
              event[required]")
+    end_event = EndEvent.Field(
+        description="Mutation to end a calendar event given the arguments\
+            \n- calendar_id: The unique identifier of the calendar event\
+            [required]\n- event_id: The unique identifier of the target\
```
TeamoreA: Good work @ThaDeveloper . However, consider deleting one of the duplicate `event_id` field description

api/events/schema.py
ThaDeveloper: @rajeman What happening here is in case the `check_event_in_db` doesn't return an event we will try to save the event and since the id doesn't exist it will throw an error.

api/events/schema.py
ThaDeveloper: @TeamoreA This has been worked on.

api/events/schema.py
rajeman: @ThaDeveloper  I need more clarification :
 if a user tries to end an event and the event is not in the db, what is the update for?

ThaDeveloper: @rajeman It's simply for validation, means they tried to end an event that doesn't exist and calling save on the event model does that. Also, refer to `CancelEvent` and `Checkin`

rajeman: @ThaDeveloper 
I get you now. The confusion is from the name of that method  `check_event_in_db`. The name doesn't indicate it also tries to end an event

PR #360, opened 2019-04-16
#165124861 Add User Location Automatically upon User Registration v1

verenceLola:
### What does this PR do?
Add user location automatically upon user registration 
Upon successful user registration, update the location table to add the location of the user if it doesn't exist in the database.
 - Checkout by running `git checkout ft-add-location-name-v1-165124861`
 - Perform any query or mutation with a valid token from a new user(not already in the database)
 - Check your users table to make sure that the user was successfully added. Now open the location table to confirm that the new user location has also been added.

Currently, when a new user is added into the database, the location of the user has to be manually added to the location table. To prevent this, upon user registration, the location of the user should be added automatically into the database.
Since the user information only has the location name and not other fields present in the location table, constraints such as NOT NULL should be removed to permit NULL values in fields such as `abbreviation`. 
[#165124861](https://www.pivotaltracker.com/story/show/165124861)


helpers/location/location.py
```diff
@@ -0,0 +1,8 @@
+from api.location.models import Location
+
+
+def check_and_add_location(location_name):
```
Makwe-O: LGTM. Nice work. Can you please add documentation for this new function

PR #361, opened 2019-04-16
#165210376 Filter meeting rooms by wings and floors

TeamoreA:
### What does this PR do?
Adds a `roomLabels` argument to `allRooms` query that filters rooms by wings and floors
When a user provides the argument `roomLabels` to the `allRooms` query the query returns all rooms that have that label. One can also provide comma separated labels to perform a deeper filter and only rooms that contain the provided labels will be returned
 - Checkout by running `gco ft-filter-meeting-rooms-by-wings-and-floors-165210376`
 - Perform the mutation below to query rooms with the specified labels.
```
{
  allRooms(roomLabels:"Block A, Wing B"){
    rooms{
      id
      name
      roomLabels
    }
  }
}
```
You must create rooms with the room labels you want to query with, otherwise, no rooms will be returned.
[#165210376](https://www.pivotaltracker.com/story/show/165210376)
### Checklist:
### Screenshots
- #### Query for all rooms with `roomLabels` filter
image

### DIFF
--- a/api/room/schema_query.py
+++ b/api/room/schema_query.py
@@ -116,6 +116,7 @@ class Query(graphene.ObjectType):
     """
     all_rooms = graphene.Field(
         PaginatedRooms,
+        room_labels=graphene.String(),
         page=graphene.Int(),
         per_page=graphene.Int(),
         capacity=graphene.Int(),
@@ -130,7 +131,8 @@ class Query(graphene.ObjectType):
             \n- resources: Resuources found in the room\
             \n- location: Location of the room\
             \n- office: Office where the room is found\
-            \n- devices: Devices that are in a room"
+            \n- devices: Devices that are in a room\
+            \n- room_labels: Labels to filter the rooms with"
     )
     get_room_by_id = graphene.Field(
         Room,
--- a/tests/test_rooms/test_room_filter.py
+++ b/tests/test_rooms/test_room_filter.py
@@ -8,7 +8,11 @@ from fixtures.room.filter_room_fixtures import (
     filter_rooms_by_location,
     filter_rooms_by_location_response,
     filter_rooms_by_location_capacity,
-    filter_rooms_by_location_capacity_response
+    filter_rooms_by_location_capacity_response,
+    filter_rooms_by_wings_and_floors,
+    filter_rooms_by_wings_and_floors_response,
+    filter_rooms_by_non_existent_room_label,
+    filter_rooms_by_non_existent_room_label_response
 )
 
 sys.path.append(os.getcwd())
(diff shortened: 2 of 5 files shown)


TeamoreA:
> Good work @TeamoreA I have a question though
> What if a room has only one of the labels in the query, shouldn't the query return this room too?
> At the moment I found it returns the room with exact labels in the query

Yeah, it returns all rooms with the specified room label if you only provide one label. Kindly test it out again with several rooms.

rajeman:
@TeamoreA I found out the tags are case sensitive. Shouldn't they be case insensitive? i.e. if I pass `block A` instead of `Block A`, I feel it should give the same result.

TeamoreA:
> @TeamoreA I found out the tags are case sensitive. Shouldn't they be case insensitive? i.e. if I pass `block A` instead of `Block A`, I feel it should give the same result.
Since the filter labels will be supplied from a frontend dropdown that will still be populated from the backend, case insensitivity wouldn't be important for this case. @rajeman

d-kahara:
> > @TeamoreA I found out the tags are case sensitive. Shouldn't they be case insensitive? i.e. if I pass `block A` instead of `Block A`, I feel it should give the same result.
> > Since the filter labels will be supplied from a frontend dropdown that will still be populated from the backend, case insensitivity wouldn't be important for this case. @rajeman

Considering that we're building an API which could possibly be integrated with other APIs, not only front-end frameworks, I think @rajeman 's observation is valid.

tests/test_rooms/test_room_filter.py
```diff
@@ -36,3 +38,10 @@ def test_filter_room_by_location_capacity(self):
             filter_rooms_by_location_capacity,
             filter_rooms_by_location_capacity_response
         )
+
+    def test_filter_room_by_wings_and_floors(self):
```
Bonifase: Write tests for all the edge cases, i.e when there is no match and no results, when the filter parameter is not valid. Also, take care of the case sensitivity

api/room/schema_query.py
```diff
@@ -115,6 +115,7 @@ class Query(graphene.ObjectType):
     """
     all_rooms = graphene.Field(
         PaginatedRooms,
+        room_labels=graphene.String(),
```
jesseinit: Good work on the PR @TeamoreA,
Kindly add room label to the description on line `126`

TeamoreA: This has been added @jesseinit

tests/test_rooms/test_room_filter.py
TeamoreA: This has been worked on @Bonifase

rajeman:
Working as expected. Well done

PR #371, opened 2019-04-23
#165014764 Update an assigned room resource

kwanj-k:
#### What does this PR do?

Update an assigned room resource

Currently, one can assign a resource to a room using an association object, this PR makes it possible to update the quantity on the association object

- git pull and checkout to the branch ft-update-assigned-resource-165014764
- run application
- Run the following mutation
```
mutation{
        updateAssignedResource(roomId:1, resourceId:1, quantity: 5){
            roomResource{
                roomId
                resourceId
                quantity
            }
        }
    }
```
[165014764](https://www.pivotaltracker.com/story/show/165014764)

### Background information
* Assign a resource first
* Ensure the resource of id one is assigned to room one to use the above mutation

### Screenshots
Screenshot 2019-04-23 at 18 30 49

### Checklist:

### DIFF
--- a/api/room_resource/schema.py
+++ b/api/room_resource/schema.py
@@ -79,6 +79,48 @@ class AssignResource(graphene.Mutation):
         return AssignResource(room_resource=room_resources)
 
 
+class UpdateAssignedResource(graphene.Mutation):
+    """
+        Update assigned resource in a room
+    """
+    class Arguments:
+        resource_id = graphene.Int(required=True)
+        room_id = graphene.Int(required=True)
+        quantity = graphene.Int(required=True)
+    room_resource = graphene.Field(RoomResource)
+
+    @Auth.user_roles('Admin')
+    def mutate(self, info, resource_id, room_id, **kwargs):
+        exact_resource = ResourceModel.query.filter_by(
+            id=resource_id).first()
+        room_resource_query = RoomResource.get_query(info)
+        room_resources = room_resource_query.filter(
+            RoomResourceModel.room_id == room_id).all()
+        if not room_resources:
+            raise GraphQLError('Room has no assigned resource')
+        current_resource = None
+        for room_resource in room_resources:
+            if room_resource.resource_id == resource_id:
+                current_resource = room_resource.quantity
+                break
+        if not current_resource:
+            raise GraphQLError('Resource does not exist in the room')
+        if kwargs['quantity'] < 0:
+            raise GraphQLError(
+                'Assigned quantity cannot be less than zero'
+            )
+        available_resources = exact_resource.quantity
+        total_resources = current_resource + available_resources
+        if kwargs['quantity'] > total_resources:
+            raise GraphQLError(
+                'Assigned resource cannot exceed available quantity'
+            )
+        exact_resource.quantity = total_resources - kwargs['quantity']
+        update_entity_fields(room_resource, **kwargs)
+        room_resource.save()
+        return UpdateAssignedResource(room_resource=room_resource)
+
+
 class PaginatedResource(Paginate):
     """
         Returns paginated room resources
@@ -253,4 +295,10 @@ class Mutation(graphene.ObjectType):
             \n- roomId: Unique key identifier of a room[required]\
             \n- resourceId: Unique key identifier of a resource[required]\
             \n- quantity: The number of resources to be assigned to a room`"
+    )
+    update_assigned_resource = UpdateAssignedResource.Field(
+        description="Updates the quantity of an assigned resource\
+            \n- room_id: The room id of the room with the resource\
+            \n- resource_id: The resource id\
+            \n- quantity: The new quantity of the resource"
     )
--- a/api/room_resource/schema.py
+++ b/api/room_resource/schema.py
@@ -94,16 +94,16 @@ class UpdateAssignedResource(graphene.Mutation):
         exact_resource = ResourceModel.query.filter_by(
             id=resource_id).first()
         room_resource_query = RoomResource.get_query(info)
-        room_resources = room_resource_query.filter(
+        rooms_with_resource = room_resource_query.filter(
             RoomResourceModel.room_id == room_id).all()
-        if not room_resources:
+        if not rooms_with_resource:
             raise GraphQLError('Room has no assigned resource')
         current_resource = None
-        for room_resource in room_resources:
+        for room_resource in rooms_with_resource:
             if room_resource.resource_id == resource_id:
                 current_resource = room_resource.quantity
                 break
-        if not current_resource:
+        if current_resource is None:
             raise GraphQLError('Resource does not exist in the room')
         if kwargs['quantity'] < 0:
             raise GraphQLError(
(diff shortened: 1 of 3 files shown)


api/room_resource/schema.py
```diff
@@ -75,6 +75,47 @@ def mutate(self, info, **kwargs):
...
+    room_resource = graphene.Field(RoomResource)
+
+    @Auth.user_roles('Admin')
+    def mutate(self, info, resource_id, room_id, **kwargs):
+        room_resource_query = RoomResource.get_query(info)
+        room = room_resource_query.filter(
+            RoomResourceModel.room_id == room_id).all()
+        if not room:
+            raise GraphQLError('Room has no assigned resource')
+        resource = [resource for resource in room if resource.resource_id == resource_id]  # noqa
```
d-kahara: For readability purposes, perhaps this could be a normal nested for loop?

api/room_resource/schema.py
```diff
@@ -75,6 +75,47 @@ def mutate(self, info, **kwargs):
...
+        room_resource = resource[0]
+        current_quantity = room_resource.quantity
+        new_quantity = kwargs['quantity']
+        if new_quantity < 0:
+            raise GraphQLError(
+                "Assigned quantity cannot be less than zero"
+            )
+        exact_resource = ResourceModel.query.filter_by(
+            id=resource_id).first()
+        unassigned_quantity = exact_resource.quantity
```
d-kahara: Could there be a comment here to explain what exactly `unassigned_quantity` is?

tests/test_room_resource/test_update_assigned_resource.py
```diff
@@ -0,0 +1,57 @@
...
+)
+from fixtures.room.assign_resource_fixture import (
+    assign_resource_mutation
+)
+sys.path.append(os.getcwd())
+
+
+class TestUpdateRoomResorce(BaseTestCase):
+
+    def test_update_assigned_resource_by_non_admin(self):
```
d-kahara: Please provide documentation for this and other tests

api/room_resource/schema.py
kwanj-k: Thanks, @d-kahara 
This has been worked on

tests/test_room_resource/test_update_assigned_resource.py
kwanj-k: This has been worked on

fixtures/room_resource/update_assigned_resource.py
```diff
@@ -0,0 +1,61 @@
+null = None
```
ThaDeveloper: Good work @kwanj-k . Why do you need this variable here?

tests/test_room_resource/test_update_assigned_resource.py
```diff
@@ -0,0 +1,76 @@
...
+    update_non_existing_room,
+    update_non_existing_resource
+)
+from fixtures.room.assign_resource_fixture import (
+    assign_resource_mutation
+)
+sys.path.append(os.getcwd())
+
+
+class TestUpdateRoomResorce(BaseTestCase):
```
ThaDeveloper: Kindly change the name of this class, there is a similar name in `test_update_room_resource.py`. If both classes are testing a related feature I would recommend putting the test cases together.

fixtures/room_resource/update_assigned_resource.py
```diff
@@ -0,0 +1,61 @@
...
+            roomResource{
+                roomId
+                resourceId
+                quantity
+            }
+        }
+    }
+'''
+
+expected_update_assigned_resource_query = {
```
ThaDeveloper: This would be better named `update_assigned_resource_query_response`

tests/base.py
```diff
@@ -179,6 +179,14 @@ def get_admin_location_id(self):
         location = Location.query.filter_by(name=user.location).first()
         return location.id
 
+    def assign_resource(self, query):
```
TeamoreA: @kwanj-k ,I suppose we don't need this method since there is a method that works the same in `CommonTestCases` class.

TeamoreA: @kwanj-k ,I suppose we don't need this method since there is a method that works the same in CommonTestCases class.

d-kahara: @TeamoreA, this is the method that was discussed during the team sync for assigning a resource to a room so that updating or querying can be tested

kwanj-k: @TeamoreA as @d-kahara has mentioned, this helper method is solely meant to just assign a resource to a room during testing, something that is not present at the moment

TeamoreA: My idea is that you can reuse the existing methods like `admin_token_assert_equal` @d-kahara .

rajeman:
@kwanj-k Can you update a resource that is not in your location?

kwanj-k:
> @kwanj-k Can you update a resource that is not in your location?

Am not updating the resources table, am only updating the room_resources table(the association object) which does not have a location column

rajeman:
> > @kwanj-k Can you update a resource that is not in your location?
> 
> Am not updating the resources table, am only updating the room_resources table(the association object) which does not have a location column

@kwanj-k Ok. Can you update a`room resource` that is not in your location?

tests/base.py
kwanj-k: Thanks for the feedback, this has been worked on @TeamoreA

api/room_resource/schema.py
```diff
@@ -75,6 +75,48 @@ def mutate(self, info, **kwargs):
...
+        Update assigned resource in a room
+    """
+    class Arguments:
+        resource_id = graphene.Int(required=True)
+        room_id = graphene.Int(required=True)
+        quantity = graphene.Int(required=True)
+    room_resource = graphene.Field(RoomResource)
+
+    @Auth.user_roles('Admin')
+    def mutate(self, info, resource_id, room_id, **kwargs):
```
verenceLola: I think we could use `**kwargs` like this to reduce the number of arguments.
```suggestion
    def mutate(self, info, **kwargs):
```

kwanj-k: @verenceLola thanks for the review, however, I need to  pass only the quantity as key-value arguments to the update entity func which will not be possible with args and kwargs

api/room_resource/schema.py
```diff
@@ -79,6 +79,48 @@ def mutate(self, info, **kwargs):
...
+        room_resource_query = RoomResource.get_query(info)
+        rooms_assigned_to_resource = room_resource_query.filter(
+            RoomResourceModel.room_id == room_id).all()
+        if not rooms_assigned_to_resource:
+            raise GraphQLError('Room has no assigned resource')
+        current_resource = None
+        for room_resource in rooms_assigned_to_resource:
+            if room_resource.resource_id == resource_id:
+                current_resource = room_resource.quantity
+                break
```
rajeman: @kwanj-k Kindly rename the variable `rooms_assigned_to_resource` as we can only assign resources to rooms and not the other way. You can use `rooms_with_resource`

kwanj-k: Done

PR #377, opened 2019-04-25
#165599195 Fix the parent_id to String in Structures model

KMaina:
#### What does this PR do
Fix the parent_id to String and ensure that it is nullable
#### Description of the task to be completed?
- Ensure that the field is a String
- Ensure that the field is nullable
[#165599195](https://www.pivotaltracker.com/story/show/165599195)
#### Checklist:


alembic/versions/f960459f3067_change_parent_id_to_string_in_the_.py
```diff
@@ -0,0 +1,36 @@
...
+down_revision = '3942aff720bb'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.alter_column('office_structures', 'parent_id',
+               existing_type=sa.INTEGER(),
+               type_=sa.VARCHAR(),
```
Bonifase: In some migration files, the type is set to String. Here is it set to VARCHAR. What are the implications of this?

KMaina: It will still set the field to a `String` as is required.

PR #353, opened 2019-04-10
#165014830 Assign resources to a room

d-kahara:
### What does this PR do?
Assign a resource to a room
### Description of the task to be completed?
An admin user should be able to assign an already created resource to a room
### How should this be manually tested?
 - Clone the repository and carry out the steps laid out [here](https://github.com/andela/mrm_api).
 - Checkout by running `git checkout ft-assign-resource-to-room-165014830`
 - Perform the mutation below to assign a resource to a room
```
mutation{
 assignResource(roomId:1,resourceId:2, quantity:9){
  roomResource{
   quantity
 }
 }
}
```

### Any background context you want to provide?
  Ensure that you run mutations for:
 - Creating a room
 - Creating a resource

### What are the relevant pivotal tracker stories?
[#165014830](https://www.pivotaltracker.com/story/show/165014830)
### Checklist:
- [x]  My code follows the style guidelines of this project
- [x] I have linted my code prior to submission
- [x] I have made the corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] New and existing unit tests pass locally with my changes
- [x] Implementation works according to expectations
### Screenshot:
Screenshot 2019-04-11 at 21 48 21

PR #378, opened 2019-04-26
#165535238 Delete assigned room resource

jesseinit:
### What does this PR do?
Enable admins to delete a resource assigned to a room

- Implement logic to aid feature functionality
- Write tests for the feature

1. Git pull branch `ft-delete-assigned-resource-165535238`
2. Follow instructions to assign a resource to a room from [#353](https://github.com/andela/mrm_api/pull/353)
2. Run the mutation below to delete the assigned resource
```
mutation {
  deleteAssignedResource(resourceId: 1, roomId: 1) {
    roomResource {
      roomId
      resourceId
      quantity
    }
  }
}
```

### Screenshots?
* On Success
Screenshot 2019-05-03 at 11 28 47 PM

* On Failure 
Screenshot 2019-05-03 at 11 29 37 PM

[#165535238](https://www.pivotaltracker.com/story/show/165535238)

### Checklist:

### DIFF
--- a/api/room_resource/schema.py
+++ b/api/room_resource/schema.py
@@ -107,6 +107,35 @@ class UpdateAssignedResource(graphene.Mutation):
         return UpdateAssignedResource(room_resource=room_resource)
 
 
+class DeleteAssignedResource(graphene.Mutation):
+    """
+        Delete assigned resource in a room
+    """
+    class Arguments:
+        resource_id = graphene.Int(required=True)
+        room_id = graphene.Int(required=True)
+    room_resource = graphene.Field(RoomResource)
+
+    @Auth.user_roles('Admin')
+    def mutate(self, info, resource_id, room_id):
+        query = RoomResource.get_query(info)
+        exact_room = RoomModel.query.filter_by(id=room_id).first()
+        if not exact_room:
+            raise GraphQLError('Room does not exist')
+        exact_resource = ResourceModel.query.filter_by(
+            id=resource_id).first()
+        if not exact_resource:
+            raise GraphQLError('Resource does not exist')
+        assigned_resource = query.filter(
+            RoomResourceModel.room_id == room_id,
+            RoomResourceModel.resource_id == resource_id).first()
+        if not assigned_resource:
+            raise GraphQLError(
+                'The resource has not been assigned to the specified room')
+        assigned_resource.delete()
+        return DeleteAssignedResource(room_resource=assigned_resource)
+
+
 class PaginatedResource(Paginate):
     """
         Returns paginated room resources
@@ -148,9 +177,9 @@ class CreateResource(graphene.Mutation):
         resource = ResourceModel(**kwargs)
         payload = {
             'model': ResourceModel, 'field': 'name', 'value':  kwargs['name']
-            }
+        }
         with SaveContextManager(
-          resource, 'Resource', payload
+            resource, 'Resource', payload
         ):
             return CreateResource(resource=resource)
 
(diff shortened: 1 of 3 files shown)


tests/base.py
```diff
@@ -99,6 +100,12 @@ def setUp(self):
             resource = Resource(name='Markers',
                                 quantity=3)
             resource.save()
+            resource_2 = Resource(name='Pen',
```
verenceLola: Nice work @jesseinit . I have a small concern though. Is it possible to just use the `Markers` resources instead of adding a new one.

api/room_resource/schema.py
```diff
@@ -79,6 +79,31 @@ def mutate(self, info, **kwargs):
...
+    room_resource = graphene.Field(RoomResource)
+
+    @Auth.user_roles('Admin')
+    def mutate(self, info, resource_id, room_id):
+        query = RoomResource.get_query(info)
+        room_resource = ResourceModel.query.filter_by(
+            id=resource_id).first()
+        assigned_resource = query.filter(
+            RoomResourceModel.room_id == room_id,
+            RoomResourceModel.resource_id == resource_id).first()
```
Makwe-O: Nice work @jesseinit. Similar to @verenceLola comment I feel there should be 2 checks. One to check if the room exists and another to check if the resource exists in the room. That way they would have more descriptive error messages

fixtures/room_resource/delete_assigned_resource_fixtures.py
```diff
@@ -0,0 +1,32 @@
...
+        deleteAssignedResource(resourceId: 10, roomId: 10) {
+            roomResource {
+                roomId
+                resourceId
+            }
+        }
+    }
+'''
+
+delete_assigned_resource_resp = {
```
ThaDeveloper: Kindly avoid abbreviations in variable names

tests/base.py
jesseinit: I tried that approach before arriving at this. The issue is that it was breaking a test case for the `assign room` feature and also another test for the `get resource list` feature. 
I'd appreciate a better way to going about it if you've got any.

api/room_resource/schema.py
jesseinit: Nice catch guys.
I felt that the check is redundant as if there is no room with such `room_id` then the resource can't be assigned in the first place.

fixtures/room_resource/delete_assigned_resource_fixtures.py
jesseinit: Would fix and update accordingly. 👍🏾

api/room_resource/schema.py
```diff
@@ -79,6 +79,31 @@ def mutate(self, info, **kwargs):
...
+        if not assigned_resource:
+            raise GraphQLError('Resource does not exist in the room')
+        room_resource.quantity = room_resource.quantity + assigned_resource.quantity  # noqa
```
TeamoreA: Good work @jesseinit . You can get rid of the `noqa` here like this `room_resource.quantity = (
            room_resource.quantity + assigned_resource.quantity)`.

api/room_resource/schema.py
kwanj-k: I think it we could be better if you raised an error `Room does not exist` before checking for the resource as its more clear if the id is not found. @jesseinit

jesseinit: @kwanj-k I specifically asked for if this check was necessary when I was clarifying the scope of the task with the TC. His argument was that the record wouldn't exist if the room with the specified id doesn't exist in the first place and that was the reason I didn't perform that check.

jesseinit:
@daniellamarr Updated.

api/room_resource/schema.py
```diff
@@ -107,6 +107,28 @@ def mutate(self, info, resource_id, room_id, **kwargs):
...
+        room_id = graphene.Int(required=True)
+    room_resource = graphene.Field(RoomResource)
+
+    @Auth.user_roles('Admin')
+    def mutate(self, info, resource_id, room_id):
+        query = RoomResource.get_query(info)
+        assigned_resource = query.filter(
+            RoomResourceModel.room_id == room_id,
+            RoomResourceModel.resource_id == resource_id).first()
+        if not assigned_resource:
```
kwanj-k: Good work @jesseinit however, wouldn't be better to do room_id check then a resource_id check separately so as you can raise a clear and specific error. In the current implementation, it always raises   `The resource has not been assigned to the specified room ` when either the room_id or resource_id does not exist.

jesseinit: Nice catch. I'll update accordingly.

PR #379, opened 2019-04-26
 #165557236 Allow admin get rooms that contain a resource

rajeman:
### What does this PR do?
 - It allows an admin to query for the rooms that contain a particular resource

- Create a query that allows an admin to get the rooms that a resource can be found after providing the `resourceId`

-  Follow the instructions [here](ur) to assign a resource to a room. 
 - Run the query below with a valid `resourceId` and you will get the list of rooms with that resource provided you are an admin
```
query {
  roomsContainingResource(resourceId: 1) {
      room {
        name
        capacity
      }
    }
  }

```

[#165557236](https://www.pivotaltracker.com/story/show/165557236)
### Checklist:

### DIFF
--- a/api/room_resource/schema.py
+++ b/api/room_resource/schema.py
@@ -237,6 +237,11 @@ class Query(graphene.ObjectType):
         room_id=graphene.Int(),
         description="Returns a list of room's resources. Accepts the argument\
             \n- room_id: Unique identifier of a room")
+    rooms_containing_resource = graphene.List(
+      RoomResource,
+      resource_id=graphene.Int(),
+      description="Returns a list of rooms containing the supplied resource id.\
+            \n- resource_id: Unique identifier of a resource")
 
     def resolve_all_resources(self, info, **kwargs):
         # Get all resources
@@ -267,6 +272,20 @@ class Query(graphene.ObjectType):
             room_resource.quantity = resource.quantity
             resources.append(room_resource)
         return RoomResources(roomResources=resources)
+
+    @Auth.user_roles('Admin')
+    def resolve_rooms_containing_resource(self, info, resource_id):
+        resource_exists = Resource.get_query(info).filter_by(
+            id=resource_id).first()
+        if not resource_exists:
+            raise GraphQLError('Resource does not exist')
+        query = RoomResource.get_query(info).join(ResourceModel).join(
+          RoomModel)
+        rooms_with_resource = query.filter(
+                RoomResourceModel.resource_id == resource_id,
+                ResourceModel.state == "active").filter(
+                  RoomModel.state == "active")
+        return rooms_with_resource
 
 
 class Mutation(graphene.ObjectType):
--- a/tests/test_room_resource/test_query_rooms_with_resource.py
+++ b/tests/test_room_resource/test_query_rooms_with_resource.py
@@ -1,0 +1,41 @@
+from tests.base import BaseTestCase, CommonTestCases
+from fixtures.room_resource.get_rooms_with_resource_fixtures import (
+    rooms_containing_resource_query,
+    rooms_containing_resource_expected_response,
+    rooms_containing_non_existent_resource_query,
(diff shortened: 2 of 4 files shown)


api/room_resource/schema.py
```diff
@@ -226,6 +231,15 @@ def resolve_get_resources_by_room_id(self, info, room_id):
             resources.append(room_resource)
         return RoomResources(roomResources=resources)
 
+    @Auth.user_roles('Admin')
+    def resolve_rooms_containing_resource(self, info, resource_id):
+        query = RoomResource.get_query(info).join(ResourceModel).join(
+          RoomModel)
+        rooms_with_resource = query.filter(
+          RoomResourceModel.resource_id == resource_id).filter(
+          ResourceModel.state == "active").filter(RoomModel.state == "active")
```
kwanj-k: Nice work @rajeman However, why did you choose to call filter twice on the same model while you could have called it once like so 
```
rooms_with_resource = query.filter(
                RoomResourceModel.resource_id == resource_id,
                ResourceModel.state == "active").filter(RoomModel.state == "active")
```

rajeman:
@d-kahara It is the expected behavior. You want to get the rooms containing the resource with id `x`. Since there is no such room(s) you get an empty list.

api/room_resource/schema.py
```diff
@@ -226,6 +231,15 @@ def resolve_get_resources_by_room_id(self, info, room_id):
             resources.append(room_resource)
         return RoomResources(roomResources=resources)
 
+    @Auth.user_roles('Admin')
+    def resolve_rooms_containing_resource(self, info, resource_id):
+        query = RoomResource.get_query(info).join(ResourceModel).join(
```
TeamoreA: Good work @rajeman . Kindly add validation when a non-existent resource id is provided

api/room_resource/schema.py
rajeman: Thanks. I will refactor as suggested.

rajeman:
@TeamoreA, @d-kahara, @verenceLola, @kwanj-k  I have made the changes.

tests/base.py
```diff
@@ -99,6 +100,12 @@ def setUp(self):
             resource = Resource(name='Markers',
                                 quantity=3)
             resource.save()
+            resource_2 = Resource(name='Pen',
```
ThaDeveloper: @rajeman You don't need to add a new resource, just use the one already added. Also, you can choose not to assign a resource here, just call the `assign resource` assert method in your test.

rajeman: @ThaDeveloper That doesn't work. If I use the one already added, the test to assign a resource to a room will fail.

tests/test_room_resource/test_query_rooms_with_resource.py
```diff
@@ -0,0 +1,19 @@
+from tests.base import BaseTestCase, CommonTestCases
+from fixtures.room_resource.get_rooms_with_resource_structure import (
+    rooms_containing_resource_query,
+    rooms_containing_resource_expected_response
+)
+
+
+class TestStructure(BaseTestCase):
```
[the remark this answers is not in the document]
rajeman: Thanks for pointing that out

tests/base.py
ThaDeveloper: @rajeman I had made the changes above and all the tests passed.

rajeman: @ThaDeveloper Can we have a sync?

rajeman: @ThaDeveloper I don't think it is a good idea to have what I am not testing for in my test file. I had sync with @jesseinit and we found out that he would also need to copy that test case into his own file if we are to use the method you are suggesting.

ThaDeveloper: @rajeman Let's sync on this

rajeman: @ThaDeveloper Kindly check the changes I made.

PR #382, opened 2019-04-29
#165675148 Fix bug on updating room resource 

Makwe-O:
#### What does this PR do?

Fix bug when admin tries to update a resource

Currently, when the Admin tries to update a resource, he can enter the quantity to be a negative number which should not be so. This PR adds a check to ensure that does not happen with the proper error message.

- git pull and checkout to the branch bg-fix-update-resource-mutation-bug-165675148
- run application
- Run the following mutation
```
mutation {
 updateRoomResource( resourceId:2 quantity:-36 name:"peach"){
   resource{
    quantity
  }
}
}
```
[165675148](https://www.pivotaltracker.com/story/show/165675148)

### Screenshots
Screenshot 2019-04-29 at 2 38 07 PM

### DIFF
--- a/api/room_resource/schema.py
+++ b/api/room_resource/schema.py
@@ -148,9 +148,9 @@ class CreateResource(graphene.Mutation):
         resource = ResourceModel(**kwargs)
         payload = {
             'model': ResourceModel, 'field': 'name', 'value':  kwargs['name']
-            }
+        }
         with SaveContextManager(
-          resource, 'Resource', payload
+            resource, 'Resource', payload
         ):
             return CreateResource(resource=resource)
 
@@ -166,13 +166,22 @@ class UpdateRoomResource(graphene.Mutation):
 
     @Auth.user_roles('Admin')
     def mutate(self, info, resource_id, **kwargs):
-        validate_empty_fields(**kwargs)
+        name = {key: value for key, value in kwargs.items() if key ==
+                'name'}
+        validate_empty_fields(**name)
         query = Resource.get_query(info)
+        quantity = kwargs.get('quantity')
         active_resources = query.filter(ResourceModel.state == "active")
         exact_resource = active_resources.filter(
             ResourceModel.id == resource_id).first()
+
         if not exact_resource:
             raise GraphQLError("Resource not found")
+        if quantity and quantity < 0:
+            raise GraphQLError(
+                'Quantity cannot be less than zero'
+            )
+
         update_entity_fields(exact_resource, **kwargs)
         exact_resource.save()
         return UpdateRoomResource(resource=exact_resource)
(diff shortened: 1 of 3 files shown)


Makwe-O:
> Greate implementation. What if the quantity is zero(0)?. As per now I'm getting this
> image
> Is it more descriptive to instead use an error message like `Quantity should be more that zero` when the user tries to to update the quantity to something equal to or less that zero?

Noted I just found out that not only does it accept negative numbers but it does not allow zero(0) which shouldn't be so. I work on updating the validation

verenceLola:
> > Greate implementation. What if the quantity is zero(0)?. As per now I'm getting this
> > image
> > Is it more descriptive to instead use an error message like `Quantity should be more that zero` when the user tries to to update the quantity to something equal to or less that zero?
> 
> Noted I just found out that not only does it accept negative numbers but it does not allow zero(0) which shouldn't be so. I work on updating the validation

Whats the reasoning behind having a resource whose quantity is 0, considering it’ll never be assigned to any room. As in its like it doesn't exist.

rajeman:
@Makwe-O I noticed that when I reduced the quantity of a resource, It did not affect the quantity of that resource in the room(s) that it has been assigned to. I mean, if we have 30 of a resource and there are 25 in a certain room called `Valhalla`.  If we update the resource to 10, I think the quantity in `Valhalla` should not be more than 10 but it remained at 25. It can also be implemented such that it will not allow reducing a resource to an amount lower than the total assigned quantity.

Makwe-O:
> @Makwe-O I noticed that when I reduced the quantity of a resource, It did not affect the quantity of that resource in the room(s) that it has been assigned to. I mean, if we have 30 of a resource and there are 25 in a certain room called `Valhalla`. If we update the resource to 10, I think the quantity in `Valhalla` should not be more than 10 but it remained at 25. It can also be implemented such that it will not allow reducing a resource to an amount lower than the total assigned quantity.

That makes sense. I have updated it to add the check

TeamoreA:
> @Makwe-O I noticed that when I reduced the quantity of a resource, It did not affect the quantity of that resource in the room(s) that it has been assigned to. I mean, if we have 30 of a resource and there are 25 in a certain room called `Valhalla`. If we update the resource to 10, I think the quantity in `Valhalla` should not be more than 10 but it remained at 25. It can also be implemented such that it will not allow reducing a resource to an amount lower than the total assigned quantity.

I suppose this is not the case @rajeman. When you have a total of `30` resources and `25` are assigned to `Valhalla` the new count in the resources table is `5`, It is possible to update it to even `0`.

rajeman:
> > @Makwe-O I noticed that when I reduced the quantity of a resource, It did not affect the quantity of that resource in the room(s) that it has been assigned to. I mean, if we have 30 of a resource and there are 25 in a certain room called `Valhalla`. If we update the resource to 10, I think the quantity in `Valhalla` should not be more than 10 but it remained at 25. It can also be implemented such that it will not allow reducing a resource to an amount lower than the total assigned quantity.
> 
> I suppose this is not the case @rajeman. When you have a total of `30` resources and `25` are assigned to `Valhalla` the new count in the resources table is `5`, It is possible to update it to even `0`.

Oh. You are right. Sorry mixed things up. @Makwe-O

Makwe-O:
Due to the recent updates on the codebase specifically the removal of the quantity column on the resources table, this PR no longer has relevance to the overall structure of the application which is why it is being closed

PR #383, opened 2019-04-30
#165727957 Refactor Analytics Helper for Getting all Events in a Room

verenceLola:
#### What does this PR do?
This PR refactors the method used for getting all events in a room.
Currently, when processing the events in for a given room, all the events in the database are processed.This was making the application to take a lot of time to process analytics requests resulting to HTTP_TIMEOUT error. This PR fixes this by processing only events that have status set to `active` and belongs to the given `room_id` before hand hence reducing the latency in processing the analytics requests.

- git pull and checkout to the branch ch-refactor-analytics-165727957
- run application
- You can make use of any analytics query. A sample query has been provided below
```
{
  bookingsAnalyticsCount(startDate:"Jan 12 2018", endDate:"Jan 12 2019"){
    bookings
    period
  }
}
```
This query should now be resolved much faster than before the fix was implemented.
[165727957](https://www.pivotaltracker.com/story/show/165727957)
Before testing this fix, ensure that you have the latest data dump for v2. 
#### After refactor
image

### DIFF
--- a/helpers/calendar/analytics_helper.py
+++ b/helpers/calendar/analytics_helper.py
@@ -111,27 +111,13 @@ class CommonAnalytics:
             - room_id - for specific room
             - event_start_time, event_end_time(Time range)
         """
-        events = EventsModel.query.all()
-        room_events = []
-        for event in events:
-            if event.room_id == room_id and not (
-                event.start_time < event_start_time or
-                    event.end_time > event_end_time):
-                room_event = {
-                    'room_id': event.room_id,
-                    'event_start_time': event.start_time,
-                    'event_end_time': event.end_time,
-                    'event_title': event.event_title,
-                    'participants': event.number_of_participants,
-                    'event_id': event.event_id,
-                    'cancelled_status': event.cancelled,
-                    'checked_in_status': event.checked_in,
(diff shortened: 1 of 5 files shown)


helpers/calendar/analytics_helper.py
```diff
@@ -111,26 +111,28 @@ def get_all_events_in_a_room(self,
...
         """
-        events = EventsModel.query.all()
+        events = EventsModel.query.filter(
+            EventsModel.room_id == room_id,
+            EventsModel.state == 'active',
+            EventsModel.end_time < event_end_time,
+            EventsModel.start_time > event_start_time
+        )
         room_events = []
         for event in events:
```
joshuaocero: Great refactor @verenceLola! Can we use the `all()` method on `filter` so we return the filter result without having to loop through it again?

verenceLola: I've been trying to get rid of the loop, a lot of areas to refactor, but I'll be done with it in a few minutes. @joshuaocero

verenceLola: All done @joshuaocero .

PR #385, opened 2019-05-03
#165767409 Fix All Locations Query 

KMaina:
#### What does this PR do
It ensures that the query to fetch all locations fetches the resources present in the room as it was before we refactored the resource logic.
#### Description of the task to be completed?
- Get all location using the query below to fetch the resource `name`, `id`, `quantity` and `room id`
```
query GetAllLocationsQuery {
   allLocations {
       id
       name
       country
       imageUrl
       rooms {
         id
         name
         imageUrl
         calendarId
         locationId
         capacity
         firebaseToken
         resources {
           resourceId
           name
           quantity
           roomId
         }
         roomTags {
           id
           name
           description
         }
       }
     }
 }
```
image

- Refactor the `CreateResource` to ensure that we do not add quantity when adding a resource to DB.
image

- Refactor the `AssignResource` and `UpdateAssignedResource` to remove the logic that checks if the quantity in DB is less than what is being assigned to a room. 
image
image

[#165767409](https://www.pivotaltracker.com/story/show/165767409)
#### Checklist:


api/room_resource/schema.py
```diff
@@ -55,17 +55,12 @@ def mutate(self, info, **kwargs):
             id=kwargs['resource_id'], state="active").first()
         if not exact_resource:
             raise GraphQLError('Resource with such id does not exist.')
-        exact_quantity = exact_resource.quantity
```
d-kahara: @joshuaocero please shed some light on why you asked @KMaina to remove these inventory checks

ThaDeveloper:
@KMaina Why are we not counter-checking the available quantity in DB for `AssignResource` and `UpdateAssignedResource`? I mean if we only have 1 marker, should we assign a room 5?

api/room_resource/schema.py
TeamoreA: Yeah. This is a bit controversial even the fact that when creating a resource we don't include the quantity of the resource @joshuaocero

kwanj-k:
> @KMaina Why are we not counter-checking the available quantity in DB for `AssignResource` and `UpdateAssignedResource`? I mean if we only have 1 marker, should we assign a room 5?

We are no longer have quantity in the resources table, hence no need to counter check anymore @ThaDeveloper

alembic/versions/ac2686b45596_refactor_resource_and_roomresource_.py
```diff
@@ -0,0 +1,30 @@
...
+# revision identifiers, used by Alembic.
+revision = 'ac2686b45596'
+down_revision = 'f960459f3067'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_column('resources', 'quantity')
```
verenceLola: Why are we dropping the quantity field of the resources table? @KMaina cc @joshuaocero

api/room_resource/schema.py
jesseinit: @TeamoreA Can you rephrase your statement above? I don't understand what you're trying to pass across.

TeamoreA: @jesseinit it was a plus one on the comment above. Asking the reason behind removing the quantity in the resources table.

joshuaocero: @d-kahara @TeamoreA my view is that converge is not an inventory system so keeping inventory is out of scope. However, the product team has the final say so we'll be syncing with the product team to seek clarity. This PR cannot wait however as it is blocking our migration to production.

ThaDeveloper:
> > @KMaina Why are we not counter-checking the available quantity in DB for `AssignResource` and `UpdateAssignedResource`? I mean if we only have 1 marker, should we assign a room 5?
> 
> We no longer have quantity in the resources table, hence no need to counter check anymore @ThaDeveloper

@kwanj-k My concern was why we are doing away with the `quantity` aspect in Resources.

PR #395, opened 2019-05-10
#165939988 Enable room update without structure id

kwanj-k:
#### What does this PR do?

Enable room update without providing the structure id

Currently, when updating the room, the mutation requires the structure id to always be provided and 
goes on to run the structure id validation which returns structure does not exist when there is no id, this
PR makes sure the validation method is only called when the structure id is in the keyword arguments 
(kwargs)

- git pull and checkout to the branch bg-room-update-165939988
- run application
- Run the following mutation to update a room
```
mutation{
    updateRoom(roomId: 5, name: "Jinja", capacity: 8,
    roomTags: [1]){
        room{
            name
            capacity
            roomType
        }
    }
}
```
[165939988](https://www.pivotaltracker.com/story/show/165939988)

### Background information
None

### Screenshots
Without structure_id
Screenshot 2019-05-10 at 20 02 09

With structure_id
Screenshot 2019-05-10 at 20 02 42

### Checklist:

### DIFF
--- a/api/room/schema.py
+++ b/api/room/schema.py
@@ -189,7 +189,8 @@ class UpdateRoom(graphene.Mutation):
     @Auth.user_roles('Admin')
     def mutate(self, info, room_id, **kwargs):
         validate_empty_fields(**kwargs)
-        validate_structure_id(**kwargs)
+        if kwargs.get('structure_id'):
+            validate_structure_id(**kwargs)
         query_room = Room.get_query(info)
         active_rooms = query_room.filter(RoomModel.state == "active")
         room = active_rooms.filter(RoomModel.id == room_id).first()
--- a/fixtures/room/room_update_fixtures.py
+++ b/fixtures/room/room_update_fixtures.py
@@ -12,7 +12,18 @@ query_update_all_fields = '''mutation{
     }
 }
 '''
-
+query_update_without_structure_id = '''mutation{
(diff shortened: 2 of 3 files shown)


api/room/schema.py
```diff
@@ -189,7 +189,8 @@ class Arguments:
     @Auth.user_roles('Admin')
     def mutate(self, info, room_id, **kwargs):
         validate_empty_fields(**kwargs)
-        validate_structure_id(**kwargs)
+        if kwargs.get('structure_id'):
```
Bonifase: If you validate for empty fields before checking for this, does it mean that you cannot update without structure id in the kwargs?

kwanj-k: When one provides an empty string you mean, am looking into that.

api/room/schema.py
```diff
@@ -188,8 +188,13 @@ class Arguments:
 
     @Auth.user_roles('Admin')
     def mutate(self, info, room_id, **kwargs):
+        if 'structure_id' in kwargs:
+            structure_id = kwargs.get('structure_id').strip()
+            if not structure_id:
+                del kwargs['structure_id']
+            else:
+                validate_structure_id(**kwargs)
```
Bonifase: Why do you validate structure_id yet you have deleted it from kwargs in line 194?

kwanj-k: I only validate if it's not an empty string if line 194 is run, line 196 will never run

PR #397, opened 2019-05-12
#165904360 Make structure_id unique

jesseinit:
### What does this PR do?
Implements a check that ensures that an office structure cannot be added twice

- Generate migration file
- Update model definition
- Extend validate_structure_id function to check for structure_id uniqueness

2. Cd into the cloned repo and checkout to `ch-make-structure-id-unique-165904360` branch
3. Run the migration to setup the database (if you haven’t)  
5. Run the mutation below twice to trigger the structure_id uniqueness check
```
mutation {
  createOfficeStructure(data: [{structureId: "851ae8b3-48dd-46b5-89bc-ca3f8111ad87", name: "Office 1", level: 1, parentId: "1", tag: "office", position: 1, locationId: 2}, {structureId: "fd236595-71aa-407f-b417-f97ebadb94a0", name: "Office 2", level: 1, parentId: "1", tag: "office", position: 2, locationId: 2}, {structureId: "27fbc541-6c2f-437a-bb38-be9af64442c3", name: "Office 3", level: 1, parentId: "1", tag: "office", position: 2, locationId: 2}]) {
    structure {
      structureId
      name
      level
      parentId
      tag
      position
      locationId
    }
  }
}
```

Hitherto this implementation, the office_structure table does not impose the unique constraint on the stucture_id field and the above mutation doesn’t check for if the structure being created already exists in the database. This led to the possibility of having structures with the same structure ids which aren’t optimal. This PR exists to eliminate this anomaly. 

### Screenshot
* Check in action
Screenshot 2019-05-12 at 11 36 58 PM

[#165904360](https://www.pivotaltracker.com/story/show/165904360)

### Checklist:


utilities/validations.py
```diff
@@ -133,6 +133,14 @@ def validate_structure_id(**kwargs):
     table
     :param structure_id
     """
+    if(kwargs.get('check_unique_structure_id')):
+        for structure in kwargs['data']:
+            existing_structure = StructureModel.query.filter_by(
```
rajeman: @jesseinit You might want to avoid using a verb `check_unique_structure_id` as a variable name

rajeman: Also, kindly explain why the flag `check_unique_structure_id` is necessary.

jesseinit: @rajeman Thanks for the feedback, can you please kindly suggest a better variable name as I felt that variable best describes what it's doing as a parameter.
Also, the variable is sort of a settings parameter that extends the functionality of the `validate_structure_id` function. So when the variable is set to `True` it simply queries the StructureModel, checks for existing structures and if any structure is found raises an exception.

api/structure/schema.py
```diff
@@ -44,6 +45,7 @@ class Arguments:
 
     @Auth.user_roles('Admin')
     def mutate(self, info, **kwargs):
+        validate_structure_id(check_unique_structure_id=True, **kwargs)
```
kwanj-k: Great work @jesseinit your PR works as described, however, do you need this argument `check_unique_structure_id=True`,?  without it and the flag for it in your validator, the PR still works as expected, kindly elaborate on this.

jesseinit: Without the check, the PR wouldn't work as expected, It'll throw an integrity error. The check is only making sure that the structures being created doesn't already exist in the database.

utilities/validations.py
```diff
@@ -133,6 +133,14 @@ def validate_structure_id(**kwargs):
...
+                structure_id=structure['structure_id']).first()
+            if existing_structure:
```
ThaDeveloper: @jesseinit Have you considered validating for the scenario where a list of structures with the same structure_id is passed? We want to return a user-friendly message other than returning the `psycopg2.IntegrityError` from the DB.

jesseinit: @ThaDeveloper Yeah that check is incorporated in the logic. If you try to run the mutation in the PR description twice you'd see that your concern has been addressed.

utilities/validations.py
Makwe-O: I think I understand where Raji is coming from. Ideally, variables should be named with nouns and functions/methods with verbs. How about just removing the 'check'. What do you think?

jesseinit: @rajeman @Makwe-O Updated

utilities/validations.py
ThaDeveloper: I don't mean running the mutation twice, I mean adding two structures with similar Ids at the same mutation. I received an error as shown here. 
image

jesseinit: @ThaDeveloper 
It has been addressed. Kindly update the branch and retry

utilities/validations.py
```diff
@@ -133,6 +133,22 @@ def validate_structure_id(**kwargs):
     table
     :param structure_id
     """
+    if(kwargs.get('unique_structure_id')):
+        structure_id_list = [structure['structure_id']
```
Bonifase: Avoid using a list comprehension, let's have a more readable code. Structure id should be validated by the ids in the structure model.

jesseinit: Updated.

utilities/validations.py
```diff
@@ -129,13 +129,29 @@ def validate_room_labels(**kwargs):
...
+            structure_id=kwargs.get('structure_id')
+        ).first()
+        if structure is None:
+            raise GraphQLError("Structure id does not exist")
+    else:
+        structure_id_list = [structure['structure_id']
+                             for structure in kwargs['data']]
+        duplicate_structures = []
+        for structure_id in structure_id_list:
+            if structure_id_list.count(structure_id) > 1:
```
Bonifase: Why don't you just return the error message if this condition is true, instead of appending the ids in the list and checking if the list has ids?

Bonifase: Also, check `[[Finishes #165904360]` in your commit message

jesseinit: Thanks. Updated.

PR #399, opened 2019-05-15
#165903102 Fix createResponse mutation

KMaina:
### What does this PR do
It ensures that the createResponse mutation works as described.
We should ensure that it should handle all the validations, and return a response object created instead of an error or blank response when performing mutations under the following conditions:
- Run create response mutation with all the required fields
image

- Run create response mutation with only the missing items being reported
image

- Run create response mutation with response type which is different from the question type
image

- Run create response mutation when trying to respond to a question whose start date is yet to commence.
image

### Background information
- Ensure you have resources in your database.
- Ensure that one room has some resources in it.
- Ensure you have question types of `check`, `input` and `rate` to respond to. 
- The mutations for the responses for each of the question types are given below. Ensure that the `questionId` value matches the type in the `Question` table.
#### Check
```
mutation{
  createResponse(responses: [{questionId:1, missingItems:[1,2]}], roomId:1) {
    response{
      id
      questionId
      roomId
      check
      }
  }
}
```
#### Rate
```
mutation{
  createResponse(responses: [{questionId:1, rate:2}], roomId:1) {
    response{
      id
      questionId
      roomId
      }
  }
}
```
#### Input
```
mutation{
 createResponse(
   responses: [{questionId:3, textArea:"Any other suggestion"}], roomId:1) {
   response{
     id
     questionId
     roomId
     }
 }
}
```
- Ensure that at least one question's start date is way in advance to try and respond to.
[#165903102](https://www.pivotaltracker.com/story/show/165903102)
### Checklist:

### DIFF
--- a/helpers/response/create_response.py
+++ b/helpers/response/create_response.py
@@ -1,51 +1,55 @@
 from datetime import datetime
 from api.response.models import Response
 from api.room_resource.models import Resource
+from api.room.models import RoomResource
 
 
 def create_response(question_type, errors, responses, **kwargs):
-    if question_type.lower() == 'rate':
-        if 'rate' in kwargs and kwargs['rate']:
-            rating = [1, 2, 3, 4, 5]
-            if kwargs['rate'] not in rating:
-                errors.append(
-                    'Please rate between 1 and 5 for question {}'.format(
-                        kwargs['question_id']))
-                return responses, errors
-            rating = Response(
-                rate=kwargs['rate'],
-                room_id=kwargs['room_id'],
-                question_id=kwargs['question_id'],
-                created_date=datetime.now())
-            rating.save()
-            responses.append(rating)
-    if question_type.lower() == 'check':
-        if 'missing_items' in kwargs and kwargs['missing_items']:
-            response = Response(
-                check=True,
-                room_id=kwargs['room_id'],
-                question_id=kwargs['question_id'],
-                created_date=datetime.now())
-            response.save()
-            for item_id in kwargs['missing_items']:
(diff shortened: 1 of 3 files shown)


tests/test_response/test_user_response.py
```diff
@@ -125,3 +126,13 @@ def test_filter_response_by_invalid_room_id(self):
             filter_question_by_invalid_room,
             filter_question_by_invalid_room_response
         )
+
+    def test_response_for_wrong_question_type(self):
+        """
+        Testing for responding to a question with the wrong question type
```
d-kahara: Consider having this docstring as something like `Test that a user cannot respond to a question with the wrong question type`

helpers/response/create_response.py
```diff
@@ -1,51 +1,55 @@
...
 
 def create_response(question_type, errors, responses, **kwargs):
-    if question_type.lower() == 'rate':
-        if 'rate' in kwargs and kwargs['rate']:
-            rating = [1, 2, 3, 4, 5]
-            if kwargs['rate'] not in rating:
-                errors.append(
-                    'Please rate between 1 and 5 for question {}'.format(
-                        kwargs['question_id']))
+    if question_type.lower() == 'rate' and 'rate' in kwargs and kwargs['rate']: # noqa
```
mnswaleh: I think `and 'rate' in kwargs` has the same truth value as `and kwargs['rate']`

rajeman: @mnswaleh Not really. `rate in kwargs` checks whether the key `rate` is in the dict `kwargs`.
`kwargs['rate']` checks if the value of the key `rate` is truthy

PR #411, opened 2019-05-23
#166208042 Indicate auto-cancellations and app-bookings

rajeman:
#### What does this PR do?

This PR ensures bookings made from the app and events that are auto-canceled are indicated.

Currently, there is no way to differentiate between events that were auto-canceled and those that were checked-in and later ended. Also, events that are booked using the app and those that were booked using the google calendar cannot be differentiated. This PR ensures these issues are resolved.

- You might need to clear your events model to enable you to see the new events after sync
- Change the calendar id of one of your rooms to `andela.com_3131313630383336383038@resource.calendar.google.com` as this calendar has bookings with the mobile app. Make sure the `next_sync_token_column` is null
- Set all rooms to `archived` except the one you just changed the calendarId
- Sync your database with the sync events mutation
```
mutation{
    syncEventData{
        message
        }
    }
```
- Check your events model for the new events
- Confirm there are events with  the `app_booking` column set to true `True`
- 
###
- Use the cancel event mutation to cancel a valid event. Note that the cancel event mutation is the same mutation used by the android device for (10 min) auto-cancellation.
- You should see the `auto_cancelled` column for that event set to True.

```
mutation {
        cancelEvent(
        calendarId:"andela.com_3630363835303531343031@resource.calendar.google.com",
        eventId:"test_id5", eventTitle:"Onboarding", numberOfParticipants: 4,
        startTime:"2018-07-10T09:00:00Z",
        endTime:"2018-07-10T09:45:00Z")
        {
            event{
                eventId
                roomId
                checkedIn
                cancelled
                room{
                    id
                    name
                    calendarId
                }
            }
        }
    }

```
[#166208042](https://www.pivotaltracker.com/story/show/166208042)

### DIFF
--- a/api/events/schema.py
+++ b/api/events/schema.py
@@ -74,7 +74,8 @@ class CancelEvent(graphene.Mutation):
                 end_time=kwargs['end_time'],
                 number_of_participants=kwargs['number_of_participants'],
                 checked_in=False,
-                cancelled=True)
+                cancelled=True,
+                auto_cancelled=True)
             event.save()
         calendar_event = get_single_calendar_event(
                                                     kwargs['calendar_id'],
@@ -150,6 +151,7 @@ def check_event_in_db(instance, info, event_check, **kwargs):
         event_id=kwargs['event_id']).scalar()
     if event and event_check == 'cancelled':
         event.cancelled = True
+        event.auto_cancelled = True
         event.save()
         return room_id, event
     elif event and event_check == 'checked_in':
--- a/helpers/calendar/events.py
+++ b/helpers/calendar/events.py
@@ -1,3 +1,4 @@
+import os
 import datetime
 import re
 
@@ -177,6 +178,11 @@ class CalendarEvents:
                     existing_event.save()
 
                 elif not event.get("status") == "cancelled":
+                    app_booking = False
+                    devices_email = os.getenv('MAIL_USERNAME')
+                    organizer = event.get("organizer")
+                    if organizer and organizer.get('email') == devices_email:
+                        app_booking = True
                     new_event = EventsModel(
                         event_id=event.get("id"),
                         recurring_event_id=event.get("recurringEventId"),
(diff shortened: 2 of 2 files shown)


api/events/schema.py
```diff
@@ -148,6 +149,7 @@ def check_event_in_db(instance, info, event_check, **kwargs):
         event_id=kwargs['event_id']).scalar()
     if event and event_check == 'cancelled':
         event.cancelled = True
+        event.auto_cancelled = True
```
Bonifase: What is the difference between a canceled event by the user from the google calendar and that one done by the mobile app after 10 minutes into the meeting?

rajeman: Event cancellation by the mobile app uses the `cancelEvent` mutation. This mutation is used only by the mobile app for `auto_cancellation`. From my understanding of the  `sync_all_events` method in  `helpers/calendar/events.py`, Events that were canceled from the calendar are updated in the DB as `archived` if they were already in the DB. If they are not already in the DB, they are not synced into the db at all.

mnswaleh: Does this mean that an event cannot be canceled using the mobile app unless it is auto-cancellation

rajeman: @mnswaleh Yes. You can only end an already started event from the mobile app. If you want to cancel an event, you will cancel it using google calendar.

api/events/schema.py
```diff
@@ -74,7 +74,8 @@ def mutate(self, info, **kwargs):
                 end_time=kwargs['end_time'],
                 number_of_participants=kwargs['number_of_participants'],
                 checked_in=False,
-                cancelled=True)
+                cancelled=True,
```
Bonifase: When we run sync event data, events with `cancelled=True` will be archived and so will events with `auto_cancelled=True` for that reason. Will this allow us to keep track of the number of auto_cancelled events?

rajeman: Events that get archived during sync are the ones that were initially in our db and then get canceled from the google calendar. Therefore all events canceled from the google calendar are archived in our db. If the event was not canceled from the google calendar it will not get archived in our db.

rajeman: Since auto-cancellation does not happen from google calendar, such events will not get archived during sync.

PR #412, opened 2019-05-27
#166203947 Block localhost requests to deployed environments

Arusey:
#### What does this PR do?
Blocks localhost requests to deployed environments
When a user makes a request to deployed environments using the localhost, they should be blocked from doing so.
3. checkout to branch `ch-block-localhost-requests-166203947`
image

#### Background context
Inspect the logic since there is no sure way of testing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,9 +36,9 @@ gcloud_setup: &gcloud_setup
   run:
     name: setup gcloud
     command: |
-      # install 
+      # install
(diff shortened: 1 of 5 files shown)


rajeman:
@Arusey  Are you sure this implementation will work?
- There is `json.loads...` in line 26 and I don't think it will work if json is not imported
- I also think adding env variables don't usually work `out of the box` with docker. You will understand my concern if you check `config.yml`

Arusey:
> @Arusey Are you sure this implementation will work?
> 
> * There is `json.loads...` in line 26 and I don't think it will work if `json` is not imported
> * I also think adding env variables don't usually work `out of the box` with docker. You will understand my concern if you check `config.yml`

Thanks for that observation.Will continue making improvements

PR #414, opened 2019-05-28
#165907624 Refactor create device mutation 

leksyib:
#### What does this PR do?
- Sets device location to location from admin's token
- Removes location field from createDevice mutation

When an admin is adding a new device, the device location field should be set to the location that is in the admin's token.

3. checkout to branch `ch-refactor-create-device-165907624`
3. run the following query:

```
mutation {
  createDevice (
    name: "samsung s3rt7",
    roomId: 1,
    deviceType: "tablet"
  ) {
    device {
      name
      location
    }
  }
}
```

#### Screenshots
image

[#165907624](https://www.pivotaltracker.com/story/show/165907624)

The `location` field would be the location in your token.

### DIFF
--- a/tests/test_devices/test_create_device.py
+++ b/tests/test_devices/test_create_device.py
@@ -2,7 +2,6 @@ from tests.base import BaseTestCase, CommonTestCases
 from fixtures.devices.devices_fixtures import (
     devices_query,
     devices_query_response,
-    create_devices_query,
     create_device_query_invalid_room
 )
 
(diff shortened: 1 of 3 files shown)


tests/test_devices/test_create_device.py
```diff
@@ -23,16 +22,6 @@ def test_device_creation(self):
         query = self.app_test.post(devices_query, headers=headers)
         self.assertEqual(query.data, devices_query_response)
 
-    def test_create_device_in_room_that_is_not_in_admin_location(self):
```
Bonifase: Could you kindly explain why you are removing these test cases?

leksyib: This test case isn't needed again since we're getting the location from admin's token. There used to be a function that returns an error if someone tries to create a device outside his country, I took the function out because we aren't accepting the `location` parameter when creating devices anymore

leksyib:
@daniellamarr  updated

tests/test_devices/test_create_device.py
leksyib: @Bonifase

leksyib:
@mnswaleh I do not think so, but I think it's not in the scope of this PR. I'd reach out to you on slack to discuss this point you just raised.

PR #419, opened 2019-05-31
#166374529 Restrict event query by location

Bonifase:
#### What does this PR do?
-This PR is to get daily events based on the location of the user making the request.

- Refactor the query to return data based on user location
- Ensure existing tests pass

- Pull this branch `restrict-query-events-by-location-166374529`
- Add rooms from various locations and run sync data mutation to sync all events
- Run the following query 
```
query{
  allEvents(startDate: "Mar 28 2019", endDate: "Apr 28 2019"){
    day
    events{
      eventId
      eventTitle
    }
  }
}
```

n/a

[#166374529](https://www.pivotaltracker.com/story/show/166374529)

### DIFF
--- a/tests/test_events/test_query_all_events.py
+++ b/tests/test_events/test_query_all_events.py
@@ -11,7 +11,7 @@ class TestEventsQuery(BaseTestCase):
         """
         Test a user can query for all events
         """
-        CommonTestCases.user_token_assert_equal(
+        CommonTestCases.admin_token_assert_equal(
             self,
             query_events,
             event_query_response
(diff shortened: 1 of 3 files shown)


tests/test_events/test_query_all_events.py
```diff
@@ -11,7 +11,7 @@ def test_query_events(self):
         """
         Test a user can query for all events
         """
-        CommonTestCases.user_token_assert_equal(
+        CommonTestCases.admin_token_assert_equal(
```
rajeman: @Bonifase I suggest  the docstring should be updated to something close  to `Test an admin can query for all events in the location`

Bonifase: Any user can query all events, the reason admin token is used is that the regular user token does not have location attribute in it in the test environment

rajeman: Got it

PR #421, opened 2019-06-01
#166285928 Enable admin to query device by name

kwanj-k:
### What does this PR do?

Enable admin to query device by name

At the moment, one can query a device by id, this PR makes it possible to query a device by name

- git pull and checkout to the branch ft-query-device-by-name-166285928
-  run application
- Run the following query
```
query{
  deviceByName(deviceName:"apple tablet"){
    id
    name
    deviceType
    room{
      name
    }
  }
}
```
[166285928](https://www.pivotaltracker.com/story/show/166285928)

### Background information
None

### Screenshots
Screenshot 2019-06-01 at 18 48 50

### DIFF
--- a/api/devices/schema.py
+++ b/api/devices/schema.py
@@ -103,6 +103,13 @@ class Query(graphene.ObjectType):
             \n- device_id: A unique identifier of the device"
     )
 
+    device_by_name = graphene.List(
+        Devices,
+        device_name=graphene.String(),
+        description="Returns device details and accepts the argument\
+            \n- device_name: The name of the device"
+    )
+
     def resolve_all_devices(self, info):
         query = Devices.get_query(info)
         location_id = admin_roles.user_location_for_analytics_view()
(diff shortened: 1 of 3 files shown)


api/devices/schema.py
```diff
@@ -123,6 +130,18 @@ def resolve_specific_device(self, info, device_id):
...
 
+    @Auth.user_roles('Admin')
+    def resolve_device_by_name(self, info, device_name):
+        query = Devices.get_query(info)
+        device_name = ''.join(device_name.split()).lower()
+        if not device_name:
+            raise GraphQLError("Please provide the device name")
+        for device in query:
+            exact_name = ''.join(device.name.split()).lower()
+            if exact_name == device_name:
```
Bonifase: Is this filtering or searching?

kwanj-k: Am going through all the devices searching for a particular device name, so in a way, it's searching, but as @mnswaleh has pointed out I will now return  a list of devices with the same implementation

api/devices/schema.py
```diff
@@ -123,6 +130,18 @@ def resolve_specific_device(self, info, device_id):
 
         return device
 
+    @Auth.user_roles('Admin')
+    def resolve_device_by_name(self, info, device_name):
+        query = Devices.get_query(info)
+        device_name = ''.join(device_name.split()).lower()
+        if not device_name:
+            raise GraphQLError("Please provide the device name")
+        for device in query:
```
jesseinit: @kwanj-k 
`for device in devices` would improve code readability.

api/devices/schema.py
Bonifase: @kwanj-k What is the difference between searching and filtering?

kwanj-k: If you use the filter,  you get a list with the particular devices, however, I would like to make sure there is no white space or difference in casing during comparison for every device.

kwanj-k: I understand what you meant now, I have made the changes and its now a search @Bonifase

PR #434, opened 2019-06-06
#166409352 Implement resource filtering by labels

jesseinit:
### What does this PR do?
Enables an admin to filter resources by labels like buildings, floors, wings.

- Add tests
- Add `resourceLabels` to query to enable filtering by building, floors or wings.

2. Cd into the cloned repo and checkout to `ft-filter-assigned-resources-by-wings-floors-building-166409352` branch
3. Run the migration to setup the database (if you haven’t)  
5. Run the query below to filter resource by labels
```
{
  allResources(resourceLabels:"Office 1"){
    resources{
      name
    }
  }
}
```

Currently, the `allResources` query returns a list of all resources in the platform. This PR introduces a filter mechanism using the labels like building, floors or wings to filter resources (usually assigned to a room) that match the filter criteria. The `resourceLabels` argument takes a string that can be comma separated.

### Screenshot
Screenshot 2019-06-06 at 7 20 05 PM

[#166409352](https://www.pivotaltracker.com/story/show/166409352)

### Checklist:

### DIFF
--- a/api/room_resource/schema.py
+++ b/api/room_resource/schema.py
@@ -1,6 +1,6 @@
 import graphene
 from graphene_sqlalchemy import SQLAlchemyObjectType
-from sqlalchemy import func, and_
+from sqlalchemy import func, and_, String, cast
 from graphql import GraphQLError
 
 from api.room_resource.models import Resource as ResourceModel
@@ -12,6 +12,7 @@ from utilities.utility import update_entity_fields
 from helpers.auth.authentication import Auth
 from helpers.auth.error_handler import SaveContextManager
 from helpers.pagination.paginate import Paginate, validate_page
+from helpers.room_filter.room_filter import room_resources_join_room
 
 
 class Resource(SQLAlchemyObjectType):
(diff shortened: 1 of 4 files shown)


api/room_resource/schema.py
```diff
@@ -147,8 +147,17 @@ def resolve_resources(self, info):
...
-        query = Resource.get_query(info)
-        active_resources = query.filter(ResourceModel.state == "active")
+        resource_labels = self.filter_data.get('resource_labels')
+        active_resources = Resource.get_query(info).filter(
+            ResourceModel.state == "active")
+        if resource_labels:
+            active_resources = active_resources.join(
+                RoomResourceModel).join(RoomModel)
+            for resourse_label in resource_labels.split(','):
+                active_resources = active_resources.filter(
```
KMaina: Good work @jesseinit, perhaps you could move the filtering logic to the `helpers/room_filter/room_filter.py` file. It contains other instances of filtering and might be better placed there.

jesseinit: Thanks.

rajeman:
@jesseinit Kindly add more context to your PR description. I had to read the PT story before I could understand what you were trying to achieve with the code.

jesseinit:
> Good work @jesseinit however, when using `ilike` it will mean if I filter by `blockb` and I have `block b` in my DB, the query will return no resources, right? should this be the case? what do you think?



> Good work @jesseinit however, when using `ilike` it will mean if I filter by `blockb` and I have `block b` in my DB, the query will return no resources, right? should this be the case? what do you think?

I reckon that the real use case of this implementation wouldn't be as you've posited on here and even at that implementing it would lead to me using nested for loops which are quite redundant.

jesseinit:
> @jesseinit Any reason why the code coverage dropped?

Working on fixing it.

rajeman:
@Jesse should it be a concern that passing an empty string as the resource label returns all the resources?
Also, it seems the state of the room is not being considered in the filtering. Resources assigned to archived rooms are being returned after filtering.

jesseinit:
@rajeman
In my opinion passing an empty string shouldn't be a a concern. In the event you think it should be, kindly provide context to same. I'll attend to your other concern.

rajeman:
> > @rajeman
> > In my opinion, passing an empty string shouldn't be a concern. In the event, you think it should be please clarify with use cases. I'll attend to your other concern.
> 
> As an update, when I query for resources that have got any state other than `active` don't get shown in the result set.

I am talking about the state of the room they are assigned to

PR #439, opened 2019-06-17
#166540045 Enable an admin user to delete a device

koitoror:
#### What does this PR do?
enables deletion of a device by an admin user

  - enable deletion of a device by admin
  - add tests

2. Setup project according to `Readme.md`
3. checkout to branch `ft-mutation-to-delete-device-166540045 `
4. on insomnia, to delete a device with device_id of 1, send the mutation
 ```
mutation {
  deleteDevice(deviceId: 1) {
    device {
      id
      name
    }
  }
}
```
[#166540045](https://www.pivotaltracker.com/story/show/166540045)

Screenshot 2019-06-17 at 09 51 10
Screenshot 2019-06-17 at 09 49 31

### CODE REVIEW

File: api/devices/schema.py [hunk 1]
```diff
@@ -156,3 +180,9 @@ class Mutation(graphene.ObjectType):
             \n- location: The location of the device\
             \n- room_id: Unique identifier of a room where the device is found\
             [required]")
+    delete_device = DeleteDevice.Field(
+        description="Updates a given device details given the arguments to delete\
```
mnswaleh: This description is not correct, I think it is supposed to `delete` not `update`, accepts only one argument `device_id` unique identifier of a `device`

koitoror: @mnswaleh : I think i am  updating the state column changing from active to archived or delete so i think the description is accurate if not close enough

koitoror: @mnswaleh : resolved

PR #460, opened 2019-07-22
CON-162 Set the state of all new devices to active by default. 

cop1fab:
#### Description
This Pull request fixes the query that returns all devices by setting all device's state to `active` by default. 

#### Context of the problem to be solved

Upon retrieving all devices, the query would return an empty list because when devices are added to our application by default their state is `NULL`. This PR sets the state of all devices to an `active` state allowing the filter query on retrieving all devices to only return active devices. 

#### Type of Change
 bug-fix
#### How should this been tested manually
Clone the repo: git clone https://github.com/andela/mrm_api.git.
Setup the project as per the README.md.
Checkout to branch bug/CON-162-fix-device-migration.
Run the following query to create a new device.

`CREATE NEW DEVICE: Sets its state to Active by default`
```
mutation {
  createDevice (
    name: "samsung s3rt7",
    roomId: 23,
    deviceType: "tablet"
  ) {
    device {
      name
      location
      state
    }
  }
}
```
image

`GET ALL DEVICES: Returns only active devices`
```
{
 allDevices {
   name,
   id,
   state,
   deviceType
 }
}
```
image

#### Checklist

#### JIRA
[CON-162](https://andela-apprenticeship.atlassian.net/browse/CON-162?atlOrigin=eyJpIjoiMTU1MTljYzhkYWYyNGMzNmJlNmMxZjZiYmZlZjJhNTUiLCJwIjoiaiJ9)


rajeman:
@cop1fab Kindly remove the dump file.

rajeman:
@cop1fab Please provide more context on the problem this PR will solve. The description only explains how you solved it.

IssaIan:
@cop1fab, this is just an enquiry. Is there a way to make the column of devices that already existed be `active`. If not,  I think this fix then beats its purpose as we'll still have to manually update them. I would like your view on this. Otherwise, kudos on a job well done. 
Thanks.

cop1fab:
> @cop1fab Please provide more context on the problem this PR will solve. The description only explains how you solved it.

Thank you for the feedback, It has been implemented.

cop1fab:
> @cop1fab, this is just an enquiry. Is there a way to make the column of devices that already existed be `active`. If not, I think this fix then beats its purpose as we'll still have to manually update them. I would like your view on this. Otherwise, kudos on a job well done.
> Thanks.

Thank you, Devices that have been added before the state was added by default are NULL, I am in a conversation with the TTL and the DevOps team to change that on the database that is on staging but if this PR merged, all devices that will be added since then will have Active as their default state.

alembic/versions/472c199d7d44_set_nullable_false_on_state_device.py
```diff
@@ -0,0 +1,32 @@
...
+revision = '472c199d7d44'
+down_revision = 'af8e4f84b552'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.alter_column('devices', 'state',
+               existing_type=postgresql.ENUM('active', 'archived', 'deleted', name='statetype'),
```
michael-basweti: For this PR to solve the problem at hand, we must set all the existing devices to a default `active`. To do that, add a comma and `server_default='active'` after `nullable=True`. This will solve the problem at hand such that when the migrations are done, all existing devices will be set to active.

IssaIan: Thanks for the clarification @michael-basweti.

cop1fab: Thank you @michael-basweti this now works perfectly.

api/devices/schema.py
```diff
@@ -143,7 +143,7 @@ def resolve_all_devices(self, info, **kwargs):
             LocationModel.id == location_id).first()
         location_name = exact_location.name.lower()
         all_devices = query.filter(
-            func.lower(DevicesModel.location) == location_name)
+            func.lower(DevicesModel.location) == location_name, DevicesModel.state == "active").all()
```
joshuaocero: @cop1fab the filter addition makes sense but why the `all()` method?

PR #470, opened 2019-07-31
CON-171 Send email to invitees when event is auto-canceled

mwinel:
### Description
This PR adds the capability to send both the organizer or host and invitees an email notification when the event is auto-canceled. The event is auto-canceled after 10 minutes of no check-in.
### Type of change
This is a new feature.
### How Has This Been Tested?

- Clone the repo: git clone https://github.com/andela/mrm_api.git.
- Setup the project as per the ` README.md `.
- Checkout to branch ` story/CON-171-send-invitees-email-when-event-is-autocancelled `.
- Create an event on your google calendar and invite a few people. While creating your event on google calendar, add the mrm-api ` converge dummy room` as your room.
- Run the following mutation to sync your new event with the database ` syncEventData `.
```
mutation {
    syncEventData{
        message
    }
}
```
- Run the following mutation to trigger event cancellation.
```
mutation {
    cancelEvent(
        calendarId:"calendar_id",
        eventId:"event_id",
        eventTitle:"demo title",
        numberOfParticipants: 2,
        startTime:"meeting_start_time",
        endTime:"meeting_end_time") {
        event {
            eventId
            roomId
            checkedIn
            cancelled
            room {
                id
                name
                calendarId
            }
        }
    }
}
```
Ensure to match your `calendarId`, `eventId`, `startTime` and `endTime` as per your event in the events table in the database.
### JIRA
[CON-171](https://andela-apprenticeship.atlassian.net/browse/CON-171)

### DIFF
--- a/helpers/email/email.py
+++ b/helpers/email/email.py
@@ -14,7 +14,7 @@ class EmailNotification:
         """
         send email notifications after a given activity has occured
         """
-        recipients = [kwargs.get('email')]
+        recipients = kwargs.get('email')
 
         email = SendEmail(
             kwargs.get('subject'), recipients,
@@ -36,7 +36,7 @@ class EmailNotification:
         send email invite for user to join converge
         """
         return EmailNotification.send_email_notification(
-            self, email=email, subject="Invitation to join Converge",
+            self, email=[email], subject="Invitation to join Converge",
             template='invite.html',
             user_name=admin, domain=Config.DOMAIN_NAME
         )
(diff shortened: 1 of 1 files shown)


helpers/email/email.py
```diff
@@ -14,7 +14,7 @@ def send_email_notification(
         """
         send email notifications after a given activity has occured
         """
-        recipients = [kwargs.get('email')]
+        recipients = kwargs.get('email')
```
rajeman: @mwinel The other functions that use the `send_email_notification` method rely on the fact that `kwargs.get('email')`  returns a string and not a list of strings. Changing it here would affect those functions. Kindly find a way around this situation.

PR #479, opened 2019-08-07
CON-212 Add timestamps fields to database records

koitoror:
#### Description
This Pull request add created_date and updated_date fields to the database all relevant database records

3. checkout to branch `story/CON-212-add-created-and-updated-timestamps-to-database-records`

- use the query below to get all questions 
```
query {
  allQuestions {
   startDate
   dateUpdated
 }
}
```
- use the mutation below to update locations

```
mutation {
        changeUserLocation(email: "daniel.kamar@andela.com", locationId: 2) {
          user {
            name
            location
          }
        }
      }
```

#### Type of Change
- [x] New feature (non-breaking change which adds functionality

#### How has this been tested
- [x] Unit tests
- [x] Integration tests

#### Checklist

#### JIRA
[CON-212](https://andela-apprenticeship.atlassian.net/browse/CON-212)
Screenshot 2019-08-12 at 07 16 40


api/devices/models.py
```diff
@@ -13,6 +13,7 @@ class Devices(Base, Utility):
     name = Column(String, nullable=False)
     device_type = Column(String, nullable=False)
     date_added = Column(DateTime, nullable=False)
+    date_updated = Column(DateTime, nullable=False)
```
vic3king: I think you need to set a default value for this field. e.g `date_updated = Column(String, default=datetime.datetime.now)`

koitoror: @vic3king : that presents issues on testing, therefore, I have opted set nullable to true (nullable=True) and my justification for this is that whoever gets tasked to update a specific record  could manipulate nullable to false (nullable=False), set a default value if any, come up with a separate update migration file to handle that scope and also fix the tests.

api/response/models.py
```diff
@@ -27,6 +27,7 @@ class Response(Base, Utility):
     question_type = Column(Enum(QuestionType), default="rate")
     response = Column(postgresql.ARRAY(sa.String()))
     created_date = Column(DateTime, nullable=False)
+    updated_date = Column(DateTime, nullable=True)
```
rajeman: @koitoror Is there any reason why this timestamp column has a different name?

koitoror: @rajeman : unless for uniformity only, let me fix!

helpers/database.py
```diff
@@ -14,5 +14,12 @@
                                          autoflush=False,
                                          bind=engine))
 
-Base = declarative_base()
-Base.query = db_session.query_property()
+DeclarativeBase = declarative_base()
+DeclarativeBase.query = db_session.query_property()
+
+
+class Base(DeclarativeBase):
```
IssaIan: You can add docstrings here @koitoror to explain what the class does.

vic3king:
> Since these columns will be on all models and all the models make use of the `Base` object, I think you can make the `Base` object an `abstract model class` that has the two columns. Also, they should have default timestamp values and should not be nullable.

@koitoror Has this been worked on?

koitoror:
> > Since these columns will be on all models and all the models make use of the `Base` object, I think you can make the `Base` object an `abstract model class` that has the two columns. Also, they should have default timestamp values and should not be nullable.
> 
> @koitoror Has this been worked on?

Yes

alembic/versions/27a555f0c123_add_create_and_update_timestamps_on_all_.py
```diff
@@ -0,0 +1,78 @@
+"""add create and update timestamps on all records
+
+Revision ID: 27a555f0c123
+Revises: 20dd40a4c970
```
rajeman: @koitoror You will need to update the revision of this alembic file as it is no longer up to date with what we have on develop.

rajeman:
@koitoror You have not made the change. Your alembic version is out of date.

alembic/versions/27a555f0c123_add_create_and_update_timestamps_on_all_.py
koitoror: Done

koitoror:
> Nice Implementation @koitoror. What is going to be done on the columns `date_added` on `devices model and `created_date`on`responses model`now that they are the same with`date_created` column

the `date_created` and `date_updated` timestamp fields that are being introduced on all database tables are strictly for watching/noting/checking the times/dates when an entry is added or and watching for when an entry is updated and have no other purpose aside of that. therefore they would co-exist, look/feel similar but their purposes are different.

PR #490, opened 2019-08-28
CON-285 Fix error when deleted user tries to set their location

vic3king:
## Description ##
 Currently, when a deleted user tries to set their location using the `setLocation` mutation we get an error, this PR handles that error and returns a descriptive error message

  - delete a user using the `deleteUser` mutation and the n try to set a location for the user.

## Type of change ##
Please select the relevant option
- [x] Bug fix(a non-breaking change which fixes an issue)
- [ ] New feature(a non-breaking change which adds functionality)
- [ ] Breaking change(fix of a feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Checklist: ##
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing tests pass locally with my changes

### JIRA
[CON-285](https://andela-apprenticeship.atlassian.net/browse/CON-285)

### DIFF
--- a/api/user/schema.py
+++ b/api/user/schema.py
@@ -241,8 +241,7 @@ class SetUserLocation(graphene.Mutation):
         logged_in_user = get_user_from_db()
         location_id = kwargs['location_id']
         query_user = User.get_query(info)
-        user = query_user.filter(
-            UserModel.state == "active", UserModel.id == logged_in_user.id).first() # noqa
+        user = query_user.filter(UserModel.id == logged_in_user.id).first()
         if user.location:
             raise GraphQLError('This user already has a location set.')
         new_location = LocationModel.query.filter_by(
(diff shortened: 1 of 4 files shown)


api/user/schema.py
```diff
@@ -242,7 +242,7 @@ def mutate(self, info, **kwargs):
         location_id = kwargs['location_id']
         query_user = User.get_query(info)
         user = query_user.filter(
-            UserModel.state == "active", UserModel.id == logged_in_user.id).first() # noqa
+            UserModel.state == "active", UserModel.id == logged_in_user.id).first()  # noqa
```
rajeman: @vic3king This check is not necessary as a user with a non-active account will not be able to use this mutation.

vic3king: True @rajeman . I'll update it.

